Issue #858, opened 2013-11-05
[2.0] A Silex specific method is called in the Router provider


fabpot:
All `register()` providers relies only on the Pimple methods, except `RoutingServiceProvider`, which uses `$app->flush()` for the `url_generator` service.

PR #862, opened 2013-11-07
[2.0] removed the flush call in the url generator service

fabpot:
As explained in #858, the flush call in the URL generator service hurts reusability and is only useful when you need to get the URL generator before the handling of the request. As this use case is an edge case, and because it can be easily documented and explained, I propose to remove the call in 2.0.


stof:
this should be explained in an upgrade guide

PR #916, opened 2014-03-04
Support optional session invalidation on logout

neeg:
Avoid session destroying on logout by adding "invalidate_session: false" in your configuration e.g.

$app['security.firewalls'] = array(
[...]
        'logout' => array(
            'logout_path' => '/user/logout',
            'target_url' => '/logout',
            'invalidate_session' => false
        ),
[...]
);


fabpot:
Thank you @neeg.

PR #938, opened 2014-04-03
SecurityServiceProvider.php

YannickGagnon:
I think we should support route names for all configurations, currently we only have the option for login_path, this way, we'd support route name for all options: login_path, check_path, logout_path and target_url. Current use case is about keeping track o locale through the URL. With the current update, it only requires that one defines the routes and bind them.

Thanks.


fabpot:
Closing as I agree with @henrikbjorn

Issue #977, opened 2014-07-01
inherit_data form validation


Kezino:
Greetings!

I am using a form type with "interit_data = true" for separating fields into groups (used for further visual rendering as tabs). Everything seems to work perfectly, except for the validation of fields inside these tabs. All errors are "bubbled" and bind to the main parent form. The field itself always returns true for the $form->get('tab_name')->get('field_name)->isValid(); whereas $form->isValid() is false. The same thing works smoothly in Symfony. Please, advice what could be wrong.

Best regards

PR #981, opened 2014-07-11
Issue #977

Kezino:
Changed ValidatorServiceProvider to provide Validator service through ValidatorBuilder. Fixes bugs with inherit_data form validations.


fabpot:
I've made a slightly more generic version in #1089

PR #986, opened 2014-07-23
Use PSR-4 for autoloading, works same as the PSR-0 configuration. Use au...

henrikbjorn:
...toload-dev for autoloading tests


jian-wu:
:+1: 
>As of 2014-10-21 PSR-0 has been marked as deprecated. PSR-4 is now recommended as an alternative.

petrgrishin:
:+1:

fabpot:
Thank you @henrikbjorn.

PR #990, opened 2014-07-23
A solution to resolve #983 (bad comportment of Translator)

### LINKED ISSUE #983: Translator not working correctly
`$app['translator']->getLocale()` returns empty,
`$app['locale']` returns default fallback language (en)

freepius:
Implementation of @henrikbjorn solution : 

Use the Translator directly from Symfony and create a EventListener that sets the Locale on the translator based on the request (like LocaleListener does).


fabpot:
Thank you @freepius.

PR #998, opened 2014-08-13
Use ValidatorBuilder for ValidatorServiceProvider

henrikbjorn:
This makes it easier to override different stuff for the validation also use different validator api's.

The failing test is from the Locale stuff.


fabpot:
see #1089

PR #1034, opened 2014-10-07
Change closures to use $app as the first parameter.

henrikbjorn:
Almost always the first parameter is $app therefor it makes little sense
to use `use` part of the closure and having PHP do context switching all
over the place.

This also makes it explicit in the service providers exactly what $app
instance is used.

### DIFF
--- a/src/Silex/Provider/DoctrineServiceProvider.php
+++ b/src/Silex/Provider/DoctrineServiceProvider.php
@@ -35,7 +35,7 @@ class DoctrineServiceProvider implements ServiceProviderInterface
             'password' => null,
         );
 
-        $app['dbs.options.initializer'] = $app->protect(function ($app) {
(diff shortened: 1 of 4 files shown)


src/Silex/Provider/DoctrineServiceProvider.php
```diff
@@ -36,7 +36,7 @@ public function register(Container $app)
             'port'     => null,
         );
 
-        $app['dbs.options.initializer'] = $app->protect(function () use ($app) {
+        $app['dbs.options.initializer'] = $app->protect(function ($app) {
```
fabpot: I would revert this one as well.

henrikbjorn:
Rebase and reverted the initializer change.

fabpot:
Thank you @henrikbjorn.

GromNaN:
:+1:

PR #1042, opened 2014-10-23
Update intro.rst

martijn80:
I followed this usage example and wasnt seeing anything. It was because my (apache) webserver wasn't rewriting to the index.php file and thus showing a 404 page when i tried /hello/world. This extra link to the webserver configuration would have saved my some time.


fabpot:
see #6ca8bf8

PR #1060, opened 2014-11-29
Introduce KernelServiceProvider in order to remove "hacks"

henrikbjorn:
This introduces a new KernelServiceProvider that is auto registered
by Silex\Application.

The main reason for including this is to remove the ugly
"dispatcher_class" hack we used to have. But since we introduced a
seperate interface for listeners to use for subscribing we can just have
a nice normal service.


fabpot:
Thank you @henrikbjorn.

PR #863, opened 2013-11-09
Add a kernel.view listener shortcut method

simensen:

PR #1062, opened 2014-12-01
View listener with wrapper

davedevelopment:
This is built on #863, it adds a wrapper enabling chaining, type hinting and simpler listeners.

### DIFF
--- a/src/Silex/ViewListenerWrapper.php
+++ b/src/Silex/ViewListenerWrapper.php
@@ -1,0 +1,88 @@
+<?php
+
+/*
+ * This file is part of the Silex framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Silex;
+
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
+
+/**
+ * Wraps view listeners.
+ *
+ * @author Dave Marshall <dave@atst.io>
+ */
+class ViewListenerWrapper
+{
+    protected $app;
+    protected $callback;
+
+    /**
+     * Constructor.
+     *
+     * @param Application $app      An Application instance
+     * @param callable    $callback
+     */
+    public function __construct(Application $app, $callback)
+    {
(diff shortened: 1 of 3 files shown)


fabpot:
I like this addition but I'm not sure about the type hinting feature. I think the `shouldRun()` method should be removed altogether.

src/Silex/ViewListenerWrapper.php
```diff
@@ -0,0 +1,88 @@
...
+        if (!$this->shouldRun($callback, $controllerResult)) {
+            return;
+        }
+
+        $response = call_user_func($callback, $controllerResult, $event->getRequest());
+
+        if ($response instanceof Response) {
+            $event->setResponse($response);
+        } else {
+            $event->setControllerResult($response);
```
GromNaN: I would reword the "extra rule" to "if the listener doesn't return anything (ie null), ..."

Crell: I don't see why it should be undocumented. It becomes effectively a pipeline anyway, just with skippable steps.  That's the same as Symfony's View events, where if you don't ->setControllerResult() then that's the equivalent of returning null.

(The alternative if you really want to pipeline it with lots of object conversion is to leverage http://www.github.com/Crell/transformer, recently released! :smile: )

Crell:
@fabot: The hinting feature is already present for exceptions. Why not do it on view as well? It's good for DX to not have to specify in each callback that it only applies to certain types of objects.  The alternative would be lots of listeners that read:

```php
$this->view(function($result, Request $request) {
  if (! $result instanceof MyClassICareAbout) {
    return;
  }

  // ... do useful things.
};
```

That's 3 lines we can very easily eliminate from every view listener, and improve self-documentation to boot.

fabpot:
@davedevelopment Can you finish this one? Fixing CS issues and adding some documentation? I think it should be part of 1.3, not master.

fabpot:
Rebasing would also help passing the tests on 1.3.

davedevelopment:
@fabpot I can crack on with this tomorrow, though just to be clear, do you want me to scrap the type hinting part? I did notice your preference, but left it as is for discussion purposes. I prefer it with, but don't mind too much.

davedevelopment:
Just opened #1161, CS fixes and targetting 1.3, docs to come.

PR #1075, opened 2014-12-14
Fix: Require phpunit/phpunit as dev dependency

localheinz:
The section running tests in about https://github.com/silexphp/Silex/blob/1.2/README.rst#tests is currently lying, and there are two options

* require users to install phpunit globally, remove the instructions on how to install dev depencencies, and change the line for running the tests to `$ phpunit` **or**
* require `phpunit/phpunit` as a dev depenceny


localheinz:
@stof 

If you like, I can go ahead and make these changes right here.

fabpot:
Closing in favor of #1078

PR #1078, opened 2014-12-15
[1.2] Fix doc about running unit tests

GromNaN:
Alternative to #1075.

I've kept the syntax using phpunit.phar as it must be the most portable way to run PHPUnit (like it is for Composer).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,12 +39,12 @@ Read the `documentation`_ for more information.
 Tests
 -----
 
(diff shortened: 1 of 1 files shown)


GromNaN:
I've remove the `php`[...]`.phar` from the readme to reflect the best practice of installing these tools globally.

GromNaN:
Rebased and squashed commits.

fabpot:
Thank you @GromNaN.

PR #1081, opened 2014-12-23
Fixes #1017 by cascading __call to mounted ControllerCollection

### LINKED ISSUE #1017: Apply before callback to a ControllerCollection which contains another ControllerCollection
I want to add before callback to a ControllerCollection which contains another ControllerCollection, like this:

    $c1 = $app['controllers_factory'];
    $c1->get("/bar", ...);
    
    $c2 = $app['controllers_factory'];
    $c2->mount("/foo", $c1);
    $c2->before(...);

but it dosn't work for route "/foo/bar/" because of this:
https://github.com/silexphp/Silex/blob/master/src/Silex/ControllerCollection.php#L164

    foreach ($this->controllers as $controller) {
        if ($controller instanceof Controller) {
            call_user_func_array(array($controller, $method), $arguments);
        }
    }

it applies to Controller, but not to ControllerCollection. Why does it work this way?

jstewmon:
all method calls that are handled by ControllerCollection::__call
will be recorded and replayed for a mounted ControllerCollection.

forward method calls to already mounted ControllerCollections
in the list of controllers.

include support for route-level middlewares using a priority
queue for the callbacks. insertion order is preserved by converting
the priority to an array and negative count as the subordinate
priority

### DIFF
--- a/src/Silex/EventListener/MiddlewareListener.php
+++ b/src/Silex/EventListener/MiddlewareListener.php
@@ -50,7 +50,7 @@ class MiddlewareListener implements EventSubscriberInterface
             return;
         }
 
-        foreach ((array) $route->getOption('_before_middlewares') as $callback) {
+        foreach (clone ($route->getOption('_before_middlewares') ?: new \SplPriorityQueue()) as $callback) {
             $ret = call_user_func($this->app['callback_resolver']->resolveCallback($callback), $request, $this->app);
             if ($ret instanceof Response) {
                 $event->setResponse($ret);
@@ -75,7 +75,7 @@ class MiddlewareListener implements EventSubscriberInterface
             return;
         }
 
-        foreach ((array) $route->getOption('_after_middlewares') as $callback) {
+        foreach (clone ($route->getOption('_after_middlewares') ?: new \SplPriorityQueue()) as $callback) {
             $response = call_user_func($this->app['callback_resolver']->resolveCallback($callback), $request, $event->getResponse(), $this->app);
             if ($response instanceof Response) {
                 $event->setResponse($response);
--- a/src/Silex/ControllerCollection.php
+++ b/src/Silex/ControllerCollection.php
@@ -42,6 +42,7 @@ class ControllerCollection
     protected $defaultRoute;
     protected $defaultController;
     protected $prefix;
+    protected $calls = array();
 
     /**
      * Constructor.
@@ -63,6 +64,11 @@ class ControllerCollection
     public function mount($prefix, ControllerCollection $controllers)
     {
         $controllers->prefix = $prefix;
+
+        foreach ($this->calls as $call) {
+            list($method, $arguments) = $call;
+            call_user_func_array(array($controllers, $method), $arguments);
+        }
 
         $this->controllers[] = $controllers;
     }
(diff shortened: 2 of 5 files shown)


src/Silex/EventListener/MiddlewareListener.php
```diff
@@ -75,7 +75,7 @@ public function onKernelResponse(FilterResponseEvent $event)
             return;
         }
 
-        foreach ((array) $route->getOption('_after_middlewares') as $callback) {
+        foreach ($route->getOption('_after_middlewares') ?: array() as $callback) {
```
thebeline: Is this functionally different?  Or personal preference?  Please elaborate.

thebeline:
A few things:

1) There are a number of things going on in this PR, and I would like to see the two parts (cascading calls, and before/after priorities) in separate PRs, as they address separate issues.

2) On one hand I like the caching of the cascading calls, but this raises a few issues in and of it's self:
* If you are going to cache, the cached calls should be called on both `mount` as well as `match`.
* Caching the calls prevents garbage collection from ever releasing references in the cache.  This could be mitigated by clearing the calls cache on flush.

src/Silex/EventListener/MiddlewareListener.php
jstewmon: SplPriorityQueue implements Iterator, but cannot be cast to array.

Prior to this patch, `$route->getOption('_before_middlewares')` could return `null` or `array`. After this patch, it can return `null` or `SplPriorityQueue`.

I changed it, so that in the `null` case, the expression results in `array()`, which is the same as it was with the cast. But in the `SplPriorityQueue` case, the expression results in `SplPriorityQueue`.

jstewmon:
> 1) There are a number of things going on in this PR, and I would like to see the two parts (cascading calls, and before/after priorities) in separate PRs, as they address separate issues.

If separate commits would get this merged, I'd separate them, but I don't really view them as separate issues b/c the problem is that ControllerCollection middleware do not follow the same semantics as Application middleware, and this commit fixes that by addressing the two pieces of that problem.

> 2) On one hand I like the caching of the cascading calls, but this raises a few issues in and of it's self:

> If you are going to cache, the cached calls should be called on both mount as well as match.

`match` already has a caching mechanism - `ControllerCollection->defaultRoute` - the defaultRoute is cloned and passed to new Controller instances created in `match`.

> Caching the calls prevents garbage collection from ever releasing references in the cache. This could be mitigated by clearing the calls cache on flush.

I think this is the case for all callbacks passed into a Silex method - nothing is ever unset to remove references to callbacks passed into any type.

stof:
there is a big issue with this change: handling 2 requests with your Silex app is not possible anymore, because the first request will empty the priority queue when iterating over it.

jstewmon:
@stof, thanks for your comment:

> there is a big issue with this change: handling 2 requests with your Silex app is not possible anymore, because the first request will empty the priority queue when iterating over it.

I wondered about this, but there are no test cases that demonstrate how it would come to pass that the list of Route middleware would be iterated multiple times. If you can provide a test case that illustrates the problem, I think there is a simple fix.

davedevelopment:
@jstewmon 

``` php
<?php

use Silex\Application;

require "vendor/autoload.php";

$app = new Application();
$app['counter'] = 0;

$app->get('/', function () use ($app) {
    return $app['counter'];
})->before(function () use ($app) {
    $app['counter'] = $app['counter'] + 1;
});

for ($i = 0; $i < 5; $i++) {
    echo $app->handle(Request::create('/'))->getContent() . "\n";
}
```

Cloning the queue before it is iterated should fix it?

jstewmon:
@davedevelopment, thanks for the example.

Cloning the queue is exactly what I had in mind. I updated the PR branch to clone and added a test to demonstrate the use case (subsequent push was to fix an unrelated linting error in ApplicationTest.php).

PR #1083, opened 2014-12-25
Inject app in function or object that handler error.

marcusesa:
Inject $app in function or object that handler error.
When we use a object like: 
```php 
$app->error(new ErrorHandler);
``` 
or a array like: 
```php 
$app->error(array('ErrorHandler', 'myMethod'));
``` 
we cant use $app inside the class. Now we can do it:
```php 
class ErrorHandler 
{
    public function __invoke(\Exception $exception, $request, $app, $code)
    {
        if ($app['debug']) {
            return;
        }

        $app['monolog']->addError(sprintf('An error occurred: %s', $exception->getMessage()));
    }
}
``` 
or
```php 
class ErrorHandler 
{
    public function myMethod(\Exception $exception, $request, $app, $code)
    {
        if ($app['debug']) {
            return;
        }

        $app['monolog']->addError(sprintf('An error occurred: %s', $exception->getMessage()));
    }
}
```

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -490,7 +490,7 @@ takes an ``Exception`` argument and returns a response::
 
     use Symfony\Component\HttpFoundation\Response;
 
-    $app->error(function (\Exception $e, $request, $code) {
+    $app->error(function (\Exception $e, $request, $app, $code) {
         return new Response('We are sorry, but something went terribly wrong.');
     });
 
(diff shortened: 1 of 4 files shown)


fabpot:
Silex rarely injects the app itself. Instead of:

```php
$app->error(new ErrorHandler);
```

use the more explicit:

```php
$app->error(new ErrorHandler($app));
```

PR #1086, opened 2015-01-06
Fixes #1068 by checking if translation file exists before adding it as a resource

### LINKED ISSUE #1068: FormServiceProvider error missing turkish translation
In the FormServiceProvider on [line 81](https://github.com/silexphp/Silex/blob/master/src/Silex/Provider/FormServiceProvider.php#L81) the translation file is added without checking if the file exists. When loading the catalogue this resource is first and it throws an exception, which stops the loading of the rest of that locale's resources.

The only way to get past this is to overwrite the entire 'form.extensions' closure.

I'm not sure what the best approach would be. Personally, I'm caching the catalogues so doing a file existence check would not be a problem for me. I'm not sure that's the best solution for everyone though.

If the file was added to the Form component this wouldn't be a problem, unfortunately I don't speak turkish :P

CarsonF:
Some locales (such as turkish) don't have translation files, which causes errors.

### DIFF
--- a/src/Silex/Provider/FormServiceProvider.php
+++ b/src/Silex/Provider/FormServiceProvider.php
@@ -78,7 +78,10 @@ class FormServiceProvider implements ServiceProviderInterface
 
                 if (isset($app['translator'])) {
                     $r = new \ReflectionClass('Symfony\Component\Form\Form');
-                    $app['translator']->addResource('xliff', dirname($r->getFilename()).'/Resources/translations/validators.'.$app['locale'].'.xlf', $app['locale'], 'validators');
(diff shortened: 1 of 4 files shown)


fabpot:
Looks good to me. Can you add a test?

CarsonF:
How are these?

fabpot:
Tests do not pass according to Travis and it seems related to your changes.

CarsonF:
Oh I didn't think about other versions of the translation component

CarsonF:
That fixed it. Looks like there's just deprecation warnings on 2.7 build.

fabpot:
Thank you @CarsonF.

PR #1087, opened 2015-01-06
Validator translation domain is now customizable from service provider

CarsonF:
So the whole validator closure doesn't have to be rewritten to change it.


stof:
What is the use case for changing it ? All builtin messages would then have to be translated again

henrikbjorn:
I believe each field can have it changed when rendered or from the type options when adding the field.

CarsonF:
I don't think it should be assumed which domain resources are add with. Those are able to be customized I think this should be as well. Maybe those should use this variable though.

CarsonF:
All the builtin translation resources use the same variable now.

fabpot:
But then again, it makes you re-translate all built-in messages. I don't see when it would be useful to change the default.

CarsonF:
I guess I don't understand how it would make you re-translate the built-in messages? They would still be added the same way.

stof:
@CarsonF but built-in translations are provided in the ``validators``. If you change the domain, you need to duplicate them in a different domain

CarsonF:
Do you mean change it after creation? I don't intent to do that. Only override it when registering the service provider.

PR #1088, opened 2015-01-07
Allow global after middlewares to return responses like route specific ones

naderman:
Handling responses returned from after middlewares is already implemented in the middleware event listener, but was missing from the application's after method. https://github.com/naderman/Silex/blob/1.2/src/Silex/EventListener/MiddlewareListener.php#L79

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -339,7 +339,12 @@ class Application extends \Pimple implements HttpKernelInterface, TerminableInte
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


naderman:
None of the failures seem to be related to this PR at all.

localheinz:
@naderman 

See https://github.com/silexphp/Silex/pull/1077#issuecomment-67052858.

davedevelopment:
:+1:

fabpot:
Thank you @naderman.

Issue #1085, opened 2015-01-06
[1.2] Broken phpunit tests


Stelian:
Hello, there are currently 6 failed and 14 warning when running the tests for the 1.2 branch. Most of them are caused by the deprecation messages and mainly the Form / Security changes.

Question is, would a PR replacing the deprecated classes with the new ones be suited or should another approach to this be taken?

davedevelopment:
@Stelian it really depends, I'm not up to date with the changes that are being made upstream. Given silex 2.0 is on the horizon, it might be time to lock down the deps on 1.2, which might mean some of these problems go away.

fabpot:
We are still working on the Symfony side to see how to limit those problems. When it's stabilized, we will be able to update Twig to take those issues into account and limit them.

Stelian:
@davedevelopment locking down deps seems the sensitive approach right now in my opinion, especially with silex 2.0 being in the works.

@fabpot Would locking down the deps for 1.2 be something acceptable for the time being? That would fix all existing issues in phpunit without affecting any of the existing code base.

stof:
@Stelian just use stable dependencies for Symfony components (either through ``minimum-stability: stable`` which is the default in Composer, or using ``prefer-stable: true`` if you change the min stability).

Stelian:
@stof  I was just referring to the default composer.json from the 1.2 branch, not an actual production issue especially since most of the output from phpunit is deprecation warnings.

stof:
I suggest changing the 1.2 branch to make it use stable versions of the dependencies when running tests to avoid that. What do you think @fabpot ?

PR #1089, opened 2015-01-08
fixed Symfony deprecations

### LINKED ISSUE #928: Document CSRF provider deprecated and new one
from https://github.com/silexphp/Silex/commit/ae4317bb414b1715eecd7694c6dc227cf4c41cc5#commitcomment-5755559

Also it took me a while to figure this out, apparently the default CSRF provider is deprecated, and it also creates a CSRF that is a function of the SESSION ID, which assumes that you even are using sessions, and also that the session ID won't change. Apparently there is a new CSRF provider from Symfony that stores the token as a session variable:

```php
/** Use a CSRF provider that does not depend on the session ID being constant. We change the session ID randomly in middleware/session_lockout.php */
$app['form.csrf_provider'] = $app->share(function ($app) {
    $storage = new Symfony\Component\Security\Csrf\TokenStorage\SessionTokenStorage($app['session']);
    return new Symfony\Component\Security\Csrf\CsrfTokenManager(null, $storage);
});
```
Should this go in the docs?

cc/ @fabpot @joshribakoff

fabpot:
This is work in progress as I need to also fix Symfony to make it happen:

 * symfony/symfony#13315
 * symfony/symfony#13316
 * symfony/symfony#13317
 * symfony/symfony#13323

Fixes #928, #1085, #981, #998

Todo:

 - [ ] update the documentation


src/Silex/Provider/FormServiceProvider.php
```diff
@@ -101,11 +104,19 @@ public function register(Application $app)
         });
 
         $app['form.csrf_provider'] = $app->share(function ($app) {
-            if (isset($app['session'])) {
-                return new SessionCsrfProvider($app['session'], $app['form.secret']);
-            }
+            if (!class_exists('Symfony\Component\Form\Extension\DataCollector\DataCollectorExtension')) {
```
stof: It would be better to use CsrfTokenManager to detect the version being used

fabpot: That does not work as the class was added in 2.3 but not used there.

davedevelopment:
:+1: 

Are we need to add some things to the CHANGELOG for this, e.g. I'm pretty sure the CsrfTokenManager behaves differently enough to warrant a little warning to people?

fabpot:
@davedevelopment Sure, but first, let me fix all the things :) Then, we will see how to update the documentation accordingly.

davedevelopment:
@fabpot absolutely, no pressure ;)

fabpot:
I'm making some very good progress. Everything is fixed now (let's see what Travis thinks) but the validation provider, which will take a bit more time on the Symfony side.

fabpot:
If we bump Symfony to 5.3.x (x > 3), I propose to do the same for Silex 1.2.

fabpot:
\o/ Done, tests pass for all versions of Symfony.

fabpot:
Remaining tasks:

 - [ ] update docs where needed (hopefully nothing)
 - [ ] add notes in the changelog

stof:
needs to be rebased to fix conflicts according to github

src/Silex/Provider/HttpCacheServiceProvider.php
```diff
@@ -46,7 +47,11 @@ public function register(Application $app)
         });
 
         $app['http_cache.esi_listener'] = $app->share(function ($app) {
-            return new EsiListener($app['http_cache.esi']);
+            if (class_exists('Symfony\Component\HttpKernel\EventListener\SurrogateListener')) {
+                return new SurrogateListener($app['http_cache.esi']);
+            } else {
```
stof: no need to use ``else`` as the ``if`` returns

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -73,9 +76,26 @@ public function register(Application $app)
         $app['security.hide_user_not_found'] = true;
 
         $app['security'] = $app->share(function ($app) {
+            // Deprecated, to be removed in 2.0
             return new SecurityContext($app['security.authentication_manager'], $app['security.access_manager']);
```
stof: this won't work in new versions of Symfony: passing the old dependencies will create other instances of the TokenStorage and AuthorizationChecker checker internally. While this is not really an issue for the AuthorizationChecker as it is a stateless service, it is an issue for the TokenStorage, because you would then have 2 different storages, and only one of them would have the right token in it.

For the new interface, the ``security`` service should be defined with its new dependencies

README.rst
```diff
@@ -27,7 +27,7 @@ The recommended way to install Silex is through `Composer`_:
 
 .. code-block:: bash
 
-    php composer.phar require silex/silex "~1.2"
+    php composer.phar require silex/silex "~1.3"
```
stof: ``composer require`` to be consistent with the Symfony documentation which always uses the global setup

src/Silex/Provider/SecurityServiceProvider.php
fabpot: fixed

composer.json
```diff
@@ -56,7 +56,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.2.x-dev"
+            "dev-master": "1.3.x-dev"
```
GromNaN: The master branch has become `2.0.x-dev`.
You should create a branch `1.3`.

fabpot:
Closing. Will be reopened on 1.3 now that 1.2.3 is out.

fabpot:
see #1097

PR #1090, opened 2015-01-08
Updated translator to access new caching features of translation component

CarsonF:
I updated Translator and TranslationServiceProvider to accept a cache directory and debug boolean.

### DIFF
--- a/src/Silex/Provider/TranslationServiceProvider.php
+++ b/src/Silex/Provider/TranslationServiceProvider.php
@@ -59,7 +59,7 @@ class TranslationServiceProvider implements ServiceProviderInterface
         });
 
         $app['translator.cache_dir'] = null;
-        $app['translator.debug'] = function() use ($app) {
+        $app['translator.debug'] = function($app) {
             return $app['debug'];
         };
         $app['translator.domains'] = array();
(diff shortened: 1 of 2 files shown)


src/Silex/Provider/TranslationServiceProvider.php
```diff
@@ -53,6 +58,8 @@ public function register(Application $app)
             return new MessageSelector();
         });
 
+        $app['translator.cache_dir'] = null;
+        $app['translator.debug'] = $app['debug'];
```
stof: this should use a lazy definition to allow it to take into account changes to the global debug mode done after the registration of the provider

src/Silex/Provider/TranslationServiceProvider.php
```diff
@@ -53,6 +58,10 @@ public function register(Application $app)
             return new MessageSelector();
         });
 
+        $app['translator.cache_dir'] = null;
+        $app['translator.debug'] = function() use ($app) {
```
stof: this should use ``function ($app)`` to be consistent with the way the translator service is defined

CarsonF: :O I didn't realize that worked

fabpot:
#1124 has been merged now.

PR #1091, opened 2015-01-15
[Minor] Remove dead code & fix phpdoc

SofHad:


src/Silex/ServiceProviderInterface.php
```diff
@@ -34,6 +34,8 @@ public function register(Application $app);
      * This method is called after all services are registered
      * and should be used for "dynamic" configuration (whenever
      * a service must be requested).
+     *
+     * @param Application $app An Application instance
```
fabpot: This should be removed as it does not add any value over the signature.

SofHad:
I removed it for register too. To harmonize both methods ?

src/Silex/ServiceProviderInterface.php
```diff
@@ -23,8 +23,6 @@
      *
      * This method should only be used to configure services and parameters.
      * It should not get services.
-     *
-     * @param Application $app An Application instance
```
henrikbjorn: This needs to stay, also it should be added to the boot method.

fabpot: No, it does not add any value, the method signature already tell you the same.

fabpot:
Thank you @SofHad.

src/Silex/ServiceProviderInterface.php
henrikbjorn: Isn' the phpdoc invalid if it contains information, but not the parameters used? I am not saying the "An application instance" should stay. Just the `@param` tag.

PR #1092, opened 2015-01-15
The app.user does not work

harikt:
I changed it to `app.security.token` to make that work. See http://stackoverflow.com/questions/27958923/silex-the-token-storage-contains-no-authentication-token

### DIFF
--- a/doc/providers/security.rst
+++ b/doc/providers/security.rst
@@ -335,7 +335,7 @@ You can check roles in Twig templates too:
 
 .. code-block:: jinja
 
-    {% if is_granted('ROLE_ADMIN') %}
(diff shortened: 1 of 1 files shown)


davedevelopment:
I think there might be another problem there, there should be an anonymous token at the very least? If you want to use `is_granted` etc, you need to be behind a firewall.

harikt:
@davedevelopment I am really new to security component. Have been spending more time figuring out the issue. Seems 5 hours :-/ . At-least this worked for me. I have tried `app.user`, but it gave me a different error saying no user.

davedevelopment:
@harikt drop in to #silex-php and I'll help you :)

harikt:
Closing the PR .

PR #1095, opened 2015-01-17
Remove a port from the default db options

rybakit:
See https://github.com/silexphp/Silex-Providers/commit/d0f1432d395d5be4d716d3c2c23fd297b815bb4f#commitcomment-9229361


fabpot:
Thank you @rybakit.

PR #1097, opened 2015-01-20
fixed Symfony deprecations

### LINKED ISSUE #928: Document CSRF provider deprecated and new one

fabpot:
Brings Symfony 2.7 compatibility to Silex 1.3.

Fixes #928, #1085, #981, #998

Remaining tasks:

 - [x] update docs where needed (hopefully nothing)
 - [x] add notes in the changelog


fabpot:
This PR is ready for me. Any more comments before the merge?

davedevelopment:
:shipit: for me, I can't see anything awry, but I can't really see the forest for the trees :)

src/Silex/Provider/FormServiceProvider.php
```diff
@@ -104,11 +107,19 @@ public function register(Application $app)
...
+                }
 
-            return new DefaultCsrfProvider($app['form.secret']);
+                return new DefaultCsrfProvider($app['form.secret']);
+            } else {
+                // Symfony 2.4+
+                $storage = isset($app['session']) ? new SessionTokenStorage($app['session']) : new NativeSessionTokenStorage();
+
+                return new CsrfTokenManager(null, $storage);
+            }
```
stof: Technically, this is a BC break for people using the ``form.csrf_provider`` on their own, because it changes the class they get.
It might be better to register the CsrfTokenManager as a new service, and providing the BC adapter in ``form.csrf_provider`` (which would trigger a deprecation warning when they get it, but Silex itself would access the new one)

mediamonks-robert:
Can these changes also be merged to master (2.0) branch?

PR #1100, opened 2015-01-22
__call() cascade on instanceof ControllerCollection.

thebeline:
I am not sure if this is expected behaviour or not, but this will not cascade `__call()` to children `ControllerCollection`s.

This PR fixes this.

See http://stackoverflow.com/questions/28098158/silex-controllercollection-not-cascading-call-to-fellow-controllercollection


davedevelopment:
Related to #1081 and #1017

thebeline:
@davedevelopment - Are there any indications or were there any discussions around why this is called only on instances of Controllers?

My only possible thought would be circular references.  Any other possibilities?

PR #1104, opened 2015-01-28
Use the constant to refer to session auth strategy

hkdobrev:
It is always better to refer to a constant with its name rather than its value.


GromNaN:
:+1: This constant [exists since Symfony 2.0](https://github.com/symfony/symfony/blob/2.0/src/Symfony/Component/Security/Http/Session/SessionAuthenticationStrategy.php)

davedevelopment:
:+1:

fabpot:
Thank you @hkdobrev.

PR #1110, opened 2015-02-17
Example of use of regular expressions when defining routes

pwaring:
Current documentation mentions that regular expressions can be used in routes but does not provide an example.

It took me a while and some Googling before I found out how to do this, so I think an example on the page which describes routing would be useful.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -152,7 +152,7 @@ early. It actually throws an exception, which you will see how to handle later
 on.
 
 You can also use regular expressions to control how variable parts of a route
-definition are matched:
+definition are matched::
 
     $app->get('/blog/{id}', function (Silex\Application $app, $id) use ($blogPosts) {
         // process blog post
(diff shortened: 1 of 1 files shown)


doc/usage.rst
```diff
@@ -151,6 +151,17 @@ When the post does not exist, you are using ``abort()`` to stop the request
 early. It actually throws an exception, which you will see how to handle later
 on.
 
+You can also use regular expressions to control how variable parts of a route
+definition are matched:
```
sstok: You need to add a double `:` like ``::` to indicate the indented section is a code block.
See https://github.com/silexphp/Silex/pull/1110/files#diff-665ffb7d19093ce3b903440e23eb6e98R169

fabpot:
This is already well documented in the same file https://github.com/pwaring/Silex/blob/1.3/doc/usage.rst#requirements

PR #1114, opened 2015-02-23
Version 1.3 is not available yet

murnieza:


GromNaN:
Tagging 1.3.0 would be even better.

fabpot:
That's wrong. The README in master is for the upcoming 1.3. Instead, I'm going to change the default version on Github to 1.2.

PR #1120, opened 2015-03-05
Web profiler docs

yamiko-ninja:
As great as the Web Profiler is it is not in the documentation or included with Silex. Even though it is not part of the main repo I would like to add some documentation about the Web Profiler to Silex.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,7 @@
         "monolog/monolog": "~1.4,>=1.4.1"
     },
     "suggest": {
+        "silex/web-profiler": "~1.0",
         "symfony/browser-kit": "~2.3",
         "symfony/css-selector": "~2.3",
         "symfony/dom-crawler": "~2.3",
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,6 @@
         "monolog/monolog": "~1.4,>=1.4.1"
     },
     "suggest": {
-        "silex/web-profiler": "~1.0",
         "symfony/browser-kit": "~2.3",
         "symfony/css-selector": "~2.3",
         "symfony/dom-crawler": "~2.3",
--- a/doc/cookbook/profiling_your_application.rst
+++ b/doc/cookbook/profiling_your_application.rst
@@ -55,7 +55,7 @@ Before registering this service provider, you must register *ServiceControllerSe
 
 .. note::
 
-    The WebProfilerServiceProvider and some its dependencies (optional or not) comes
+    The WebProfilerServiceProvider and some of its dependencies (optional or not) come
     with the "fat" Silex archive but not with the regular one. If you are using
     Composer, add them as a dependency:
 
--- a/doc/cookbook/profiling_your_application.rst
+++ b/doc/cookbook/profiling_your_application.rst
@@ -1,5 +1,5 @@
-WebProfilerServiceProvider
-=======================
+Profiling Your Applications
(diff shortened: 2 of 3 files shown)


davedevelopment:
Perhaps it should be a cookbook entry as it's not part of the main repo?

lplume:
a cookbook about using the profiler would be awesome!

stof:
This should be documented in the provider repo, not in Silex itself IMO

yamiko-ninja:
I will switch it to a cook book and make a pull request for the provider. 

Any chance we can include the Web Profiler with Silex for future releases?

fabpot:
Including the web profiler with Silex is indeed possible, but we need to stop somewhere in the number of things we ship by default. And as the web profiler is not needed by everyone, I think it should be included by default (think of people doing only web services with Silex, which is one of the biggest usage for Silex.) With Composer, I should be easy to integrate the web profiler and having a cookbook in the official documentation will help a lot as well.

yamiko-ninja:
The profiler is the only provider that is not included that I would like added in. It is the only provider that has its own repo under (silexphp) [https://github.com/silexphp]. Ill add it as a suggestion in composer.

topikito:
I don't think it should be added by default. Its pretty easy and fast to just include it in your composer file + add the few lines it requires to work.

yamiko-ninja:
Since it shouldn't be a requirement I added it as a suggestion. Is this OK?

composer.json
```diff
@@ -46,6 +46,7 @@
         "monolog/monolog": "~1.4,>=1.4.1"
     },
     "suggest": {
+        "silex/web-profiler": "~1.0",
```
stof: -1 for this. Displaying a suggestion to anyone using Silex for this is a bad idea IMO. Suggestions are not the place to list additional providers. https://github.com/silexphp/Silex/wiki/Third-Party-ServiceProviders is this place

Suggestions are used here for packages which allow you to use more features from the silex/silex package itself (optional dependencies of the repo).

stof:
@yamiko-ninja I would rather split more providers out of the core rather than including new ones.

It would allow a better management of dependencies btw. With optional dependencies, you rely on your user to tell composer what the compatible version of the dependency is, which means that any bump of the min requirement is technically a BC break (because you don't control the requirement)

yamiko-ninja:
@stof I removed the suggestion. I am all for splitting them up if it meant better dependency management but thats a different conversation.

Ill leave i as documentation in the cookbook.

GromNaN:
Adding the suggestion was a good idea!

yamiko-ninja:
@fabpot do you think I should include it as a suggestion, requirement or not at all?

doc/cookbook/profiling_your_application.rst
```diff
@@ -0,0 +1,170 @@
...
+
+    WebProfilerServiceProvider does not come with silex. If you are using Composer, add it as a dependency:
+
+    .. code-block:: bash
+
+        composer require silex/web-profiler
+
+.. note::
+
+    The WebProfilerServiceProvider and some its dependencies (optional or not) comes
```
mTorres: The WebProfilerServiceProvider and some **of** its dependencies (optional or not) **come**

fabpot:
I've added a note about the web profiler in the docs: see 22d47c8

PR #1121, opened 2015-03-05
Add documentation about lock mode settings.

rajibahmed:
PdoSessionHandler has lock_mode settings which is really critical
for multi-server enviroment while using silex with DoctrineOrm.
DoctrineOrm uses transactional queries and PdoSessionHandler defaults
to transactional queries as well. Which creates a database leve error
notifiying transaction already exists. Changing the lock_mode to
other setting then the default one solves the issue.

### DIFF
--- a/doc/cookbook/session_storage.rst
+++ b/doc/cookbook/session_storage.rst
@@ -5,7 +5,9 @@ By default, the :doc:`SessionServiceProvider </providers/session>` writes
 session information in files using Symfony2 NativeFileSessionStorage. Most
 medium to large websites use a database to store sessions instead of files,
 because databases are easier to use and scale in a multi-webserver
-environment.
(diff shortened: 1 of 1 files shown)


stof:
@rajibahmed the PDO session handler is designed to use its own connection, not to reuse the PDO connection of the ORM

davedevelopment:
@stof that needs documenting somewhere then.

stof:
@davedevelopment please open an issue on the symfony-docs repository to ask for an update of the documentation of the PdoSessionHandler

stof:
and you could also add a note in the Silex doc in the page you are editing here

fabpot:
Closing as this is really something that needs to be done in the Symfony docs instead. As mentioned by @stof, this is not as simple as that.

PR #1122, opened 2015-03-06
Add OPTIONS request routing directly to Application

jimdoescode:
I think it's important to add this HTTP Method directly to the application because it's often called automatically by browsers during AJAX requests. Meaning it's required that anyone sending AJAX requests to Silex will have to implement some kind of OPTIONS handling. I realize that you can use the match and method calls to do this same thing but those make routing OPTIONS requests less obvious.

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -225,6 +225,19 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
     }
 
     /**
+     * Maps an OPTIONS request to a callable.
+     *
+     * @param string $pattern Matched route pattern
+     * @param mixed  $to      Callback that returns the response when matched
+     *
+     * @return Controller
+     */
+    public function options($pattern, $to = null)
+    {
+        return $this['controllers']->options($pattern, $to);
+    }
+
+    /**
      * Maps a PATCH request to a callable.
      *
      * @param string $pattern Matched route pattern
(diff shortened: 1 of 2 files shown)


jimdoescode:
I'm not sure what standards this is failing. @fabpot is there some kind of weird whitespace rule? I double checked and there are no extra spaces on those blank lines.

jimdoescode:
Strange. I don't know how that whitespace got added I never touched those lines. My keyboard must have a mind of its own.

GromNaN:
`OPTIONS` has a very specific use-case. You can use [stack-cors](https://github.com/asm89/stack-cors) to handle this method properly.

jimdoescode:
Right, it has a very specific use case but not as specific a response structure. I'd rather manually handle OPTIONS requests because then I can add any optional headers, or even a response, based on the specific route requested. Middleware (like stack-cors) wouldn't give as much flexibility as defining independent routes would.

fabpot:
Not sure if we want to add this to core. I mean, usage is not that widespread and you can easily do the same with the current code easily.

jimdoescode:
It surprises me that usage wouldn't be that widespread considering the surge in JavaScript web apps over the past few years. Some browsers like Chrome wont even complete an AJAX request until an OPTIONS request is successfully handled. 

But I guess it's not too hard to implement separately and if someone does want the ability to do that they can extend Application or switch to [Slim Framework for it.](http://docs.slimframework.com/routing/options/)

Should I close this @fabpot?

fabpot:
@jimdoescode Indeed, that's a good point.

fabpot:
Thank you @jimdoescode.

fabpot:
Merged into the upcoming 1.3 release.

fabpot:
docs updated at 77f308c

PR #1123, opened 2015-03-08
Update testing.rst

yourwebmaker:
Actually there is no "Reusing applications" topic on Usage chapter.


fabpot:
Thank you @yourwebmaker.

PR #1137, opened 2015-04-07
Correct exception message when mounting

SpacePossum:
Use a better exception message when calling `mount` with a `ControllerProviderInterface` that doesn't return an `ControllerCollection` on the `connect` call.

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -500,8 +500,7 @@ class Application extends \Pimple implements HttpKernelInterface, TerminableInte
                 throw new \LogicException(sprintf('The method "%s::connect" must return a "ControllerCollection" instance. Got: "%s', get_class($controllers), is_object($connectedControllers) ? get_class($connectedControllers) : gettype($connectedControllers)));
             }
 
-			$controllers = $connectedControllers;
-
+            $controllers = $connectedControllers;
         } elseif (!$controllers instanceof ControllerCollection) {
             throw new \LogicException('The "mount" method takes either a "ControllerCollection" or a "ControllerProviderInterface" instance.');
         }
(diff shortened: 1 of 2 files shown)


src/Silex/Application.php
```diff
@@ -495,9 +495,11 @@ public function mount($prefix, $controllers)
     {
         if ($controllers instanceof ControllerProviderInterface) {
             $controllers = $controllers->connect($this);
-        }
 
-        if (!$controllers instanceof ControllerCollection) {
+            if (!$controllers instanceof ControllerCollection) {
+                throw new \LogicException('The ControllerProviderInterface must return a ControllerCollection instance for the "connect" method.');
```
GromNaN: It would be even more explicit if we could get the actual class name.

```
The method "MyControllerProvider::connect" must return a "ControllerCollection"
```

SpacePossum: :+1:  Thanks for the suggestion, I've update the PR.

src/Silex/Application.php
```diff
@@ -494,11 +494,16 @@ public function sendFile($file, $status = 200, array $headers = array(), $conten
...
         if ($controllers instanceof ControllerProviderInterface) {
-            $controllers = $controllers->connect($this);
-        }
+            $connectedControllers = $controllers->connect($this);
+
+            if (!$connectedControllers instanceof ControllerCollection) {
+                throw new \LogicException(sprintf('The method "%s::connect" must return a "ControllerCollection" instance. Got: "%s', get_class($controllers), is_object($connectedControllers) ? get_class($connectedControllers) : gettype($connectedControllers)));
+            }
+
+			$controllers = $connectedControllers;
```
GromNaN: You used tabs instead of spaces for indentation.

src/Silex/Application.php
GromNaN: Nice :)

src/Silex/Application.php
SpacePossum: thanks for the heads up, corrected

GromNaN:
:+1:
To be complete, you can had a test that check the exception thrown.

SpacePossum:
cool, will do :)

fabpot:
Thank you @SpacePossum.

PR #1139, opened 2015-04-10
PimpleAwareEventDispatcher

Atriedes:
make current event dispatcher aware of service in pimple.

It provide similiar feature like this : http://symfony.com/doc/current/components/event_dispatcher/container_aware_dispatcher.html

this code adapted from : https://github.com/symfony/EventDispatcher/blob/master/ContainerAwareEventDispatcher.php

### DIFF
--- a/src/Silex/PimpleAwareEventDispatcher.php
+++ b/src/Silex/PimpleAwareEventDispatcher.php
@@ -1,6 +1,5 @@
 <?php
 
-
 namespace Silex;
 
 use Symfony\Component\EventDispatcher\EventDispatcher;
@@ -37,11 +36,11 @@ final class PimpleAwareEventDispatcher extends EventDispatcher
     }
 
     /**
-     * @param $event_name
-     * @param $callback
+     * @param string $event_name
+     * @param array $callback
      * @param int $priority
      */
-    public function addListenerService($event_name, $callback, $priority = 0)
+    public function addListenerService($event_name, array $callback, $priority = 0)
     {
         if (!is_array($callback) || count($callback) !== 2) {
             throw new \InvalidArgumentException("Expected an array('service', 'method') argument");
@@ -119,8 +118,8 @@ final class PimpleAwareEventDispatcher extends EventDispatcher
     }
 
     /**
-     * @param $service_id
-     * @param $class
+     * @param string $service_id
+     * @param string $class
      */
     public function addSubscriberService($service_id, $class)
     {
@@ -152,7 +151,7 @@ final class PimpleAwareEventDispatcher extends EventDispatcher
     }
 
     /**
-     * @param $event_name
+     * @param string $event_name
      */
     private function lazyLoad($event_name)
     {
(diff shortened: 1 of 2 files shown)


dosten:
Why `PimpleAwareEventDispatcher` and not `ContainerAwareEventDispatcher`?

src/Silex/PimpleAwareEventDispatcher.php
```diff
@@ -0,0 +1,178 @@
+<?php
+
+
```
dosten: Extra line

src/Silex/PimpleAwareEventDispatcher.php
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * @param Container $container
+     */
+    public function __construct(Container $container)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * @param $event_name
```
dosten: Missing param type

src/Silex/PimpleAwareEventDispatcher.php
```diff
@@ -0,0 +1,178 @@
...
+     * @param $callback
```
dosten: Missing param type

src/Silex/PimpleAwareEventDispatcher.php
```diff
@@ -0,0 +1,178 @@
...
+        } else {
+            $this->lazyLoad($event_name);
+        }
+
+        return parent::getListeners($event_name);
+    }
+
+    /**
+     * @param $service_id
+     * @param $class
```
dosten: Same here

src/Silex/PimpleAwareEventDispatcher.php
```diff
@@ -0,0 +1,178 @@
...
+     */
+    public function addSubscriberService($service_id, $class)
```
dosten: You need to type hint the `$class` parameter with `Symfony\Component\EventDispatcher\EventSubscriberInterface`

Atriedes:
@dosten : why not ContainerAwareEventDispatcher because it only implement container which is pimple, as you see in here https://github.com/symfony/EventDispatcher/blob/master/ContainerAwareEventDispatcher.php it also implement specific container. So I decided to use PimpleAware to avoid confusion.

And thanks for code review, will updated soon :)

src/Silex/PimpleAwareEventDispatcher.php
Atriedes: `$class` must be in a string that contain class name in order to perform lazy load for EvenSubscriber.

please refer to this docs : http://symfony.com/doc/current/components/event_dispatcher/container_aware_dispatcher.html#adding-subscriber-services

fabpot:
I'm :-1: on adding so much code to Silex. I'm not even sure if it's worth it.

alexantr:
@fabpot
> I'm :-1: on adding so much code to Silex.

:+1:

PR #1141, opened 2015-04-10
[Tests] Fix tests with Symfony dependencies version 2.7

petrgrishin:


fabpot:
Thanks for the pull request. I'm closing it as I've done a slightly different fix in #1142 (which involved a fix in Symfony as well).

petrgrishin:
Yes, you're right

PR #1154, opened 2015-04-23
Have `Controller::generateRouteName()` always put the method first.

LawnGnome:
At present, mounted controllers can get weird route names. For instance, let's say you have this controller definition:

    $otherController = $app['controllers_factory'];
    $otherController->get('/{name}', function (Request $request, $name) use ($app) {
      return new Response("Goodbye $name!\n", 200, ['Content-Type' => 'text/plain']);
      });
    $app->mount('/goodbye', $otherController);

The generated route name in this case will be `_goodbyeGET_name`, which technically contains everything, but is ugly.

With this PR, the route name will change to `GET_goodbye_name`, which is considerably easier to parse when debugging.

### DIFF
--- a/src/Silex/Controller.php
+++ b/src/Silex/Controller.php
@@ -109,8 +109,6 @@ class Controller
     {
         $requirements = $this->route->getRequirements();
 
-        // The trailing underscore is required to separate the method from the
-        // prefix, if one is set.
         $method = isset($requirements['_method']) ? $requirements['_method'].'_' : '';
 
         $routeName = $method.$prefix.$this->route->getPath();
@@ -118,7 +116,7 @@ class Controller
         $routeName = preg_replace('/[^a-z0-9A-Z_.]+/', '', $routeName);
 
         // Collapse consecutive underscores down into a single underscore.
-        $routeName = preg_replace('/_{2,}/', '_', $routeName);
+        $routeName = preg_replace('/_+/', '_', $routeName);
 
         return $routeName;
     }
--- a/src/Silex/Controller.php
+++ b/src/Silex/Controller.php
@@ -107,11 +107,17 @@ class Controller
     public function generateRouteName($prefix)
     {
         $requirements = $this->route->getRequirements();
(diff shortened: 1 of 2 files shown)


src/Silex/Controller.php
```diff
@@ -107,12 +108,18 @@ public function freeze()
     public function generateRouteName($prefix)
     {
         $requirements = $this->route->getRequirements();
-        $method = isset($requirements['_method']) ? $requirements['_method'] : '';
 
-        $routeName = $prefix.$method.$this->route->getPath();
+        // The trailing underscore is required to separate the method from the
+        // prefix, if one is set.
```
fabpot: I would remove the comment here.

src/Silex/Controller.php
```diff
@@ -107,12 +108,18 @@ public function freeze()
...
+        $method = isset($requirements['_method']) ? $requirements['_method'].'_' : '';
+
+        $routeName = $method.$prefix.$this->route->getPath();
         $routeName = str_replace(array('/', ':', '|', '-'), '_', $routeName);
         $routeName = preg_replace('/[^a-z0-9A-Z_.]+/', '', $routeName);
 
+        // Collapse consecutive underscores down into a single underscore.
+        $routeName = preg_replace('/_{2,}/', '_', $routeName);
```
fabpot: `_+` instead of `_{2,}`

src/Silex/Controller.php
LawnGnome: Done.

src/Silex/Controller.php
LawnGnome: Done — I wrote the regex that way to be clear about what it's doing, but the comment really covers that anyway.

fabpot:
Looks good now. But I've just realized that you worked on 1.2, but at this is a small BC break, I would prefer to merge this into the 1.3 branch. Can you create a new PR for 1.3 instead? Thanks.

LawnGnome:
Sure thing. I'll close this off and reference it in the new PR.

PR #1155, opened 2015-04-24
PHPDocs cleans ups

SpacePossum:
And some tabs versus spaces.

### DIFF
--- a/src/Silex/ControllerCollection.php
+++ b/src/Silex/ControllerCollection.php
@@ -43,9 +43,6 @@ class ControllerCollection
     protected $defaultController;
     protected $prefix;
 
-    /**
-     * Constructor.
-     */
     public function __construct(Route $defaultRoute)
     {
         $this->defaultRoute = $defaultRoute;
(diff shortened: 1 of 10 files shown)


src/Silex/ServiceProviderInterface.php
```diff
@@ -23,6 +23,8 @@
      *
      * This method should only be used to configure services and parameters.
      * It should not get services.
+     *
+     * @param Application $app
```
fabpot: This should be removed as it does not bring anything useful.

src/Silex/ServiceProviderInterface.php
```diff
@@ -32,6 +34,8 @@ public function register(Application $app);
      * This method is called after all services are registered
      * and should be used for "dynamic" configuration (whenever
      * a service must be requested).
+     *
+     * @param Application $app
```
fabpot: Same as above

fabpot:
Looks good to me apart from the 2 comments I've just made. Also, can you rebase the PR? (rebasing will remove the merge commit that we don't want when merging). Thanks.

SpacePossum:
Sure thing :) 
I've removed the two lines from the PR, rebased and commits squash'ed.

fabpot:
Thank you @SpacePossum.

PR #1157, opened 2015-04-28
auto inject controller argument

Atriedes:
Automatically inject controller argument where parameter typehinted and object exist in container.

~~~php
public function indexAction(Request $request, Foo $foo) 
{

}
~~~

In order autoinject Foo, it must have defined in container

~~~php
// service ID must full classname of object
$app[Foo:class] = function () {
    return new Foo();
}
~~~

### DIFF
--- a/src/Silex/ControllerResolver.php
+++ b/src/Silex/ControllerResolver.php
@@ -44,6 +44,10 @@ class ControllerResolver extends BaseControllerResolver
                 $request->attributes->set($param->getName(), $this->app);
 
                 break;
+            } elseif ($param->getClass() && isset($this->app[$param->getClass()->getName()])) {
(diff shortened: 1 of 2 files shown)


indeyets:
seems like a service-abuse to me. I don't like part where you name service using FQN of the class

Atriedes:
would you please explain "service-abuse" part? I don't want inject App object to my controller because it make App as Service Container that hide dependency.

Regarding FQN as service name, because it was easier way to do it :D

davedevelopment:
I don't think this is abusive, but it's definitely opinionated, something that Silex usually shies away from.

fabpot:
Indeed, this probably does not belong to Silex core. Also because in Symfony, arguments of the controllers can be determined based on some converters that takes the Request and tries to "convert" it to some objects.

PR #1158, opened 2015-04-30
Fix return of renderView to string

bburnichon:


davedevelopment:
:+1:

fabpot:
Thank you @bburnichon.

PR #1159, opened 2015-05-01
Have `Controller::generateRouteName()` always put the method first.

LawnGnome:
(As requested, this is #1154 rebased against 1.3.)

At present, mounted controllers can get weird route names. For instance, let's say you have this controller definition:

    $otherController = $app['controllers_factory'];
    $otherController->get('/{name}', function (Request $request, $name) use ($app) {
      return new Response("Goodbye $name!\n", 200, ['Content-Type' => 'text/plain']);
      });
    $app->mount('/goodbye', $otherController);

The generated route name in this case will be `_goodbyeGET_name`, which technically contains everything, but is ugly.

With this PR, the route name will change to `GET_goodbye_name`, which is considerably easier to parse when debugging.

### DIFF
--- a/src/Silex/Controller.php
+++ b/src/Silex/Controller.php
@@ -107,11 +107,14 @@ class Controller
 
     public function generateRouteName($prefix)
     {
-        $methods = implode('_', $this->route->getMethods());
+        $methods = implode('_', $this->route->getMethods()).'_';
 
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @LawnGnome.

PR #1161, opened 2015-05-05
View Listener

davedevelopment:
Previously #1062 and #863

TODO

- [x] Documentation

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -539,6 +539,58 @@ early::
         return new Response(...);
     });
 
+View Handlers
+-------------
+
+View Handlers allow you to intercept a controller result and transform it before
+it gets returned to the kernel.
+
+To register a view handler, pass a callable (or string that can be resolved to a
+callable) to the view method. The callable should accept some sort of result
+from the controller::
+
+    $app->view(function (Response $response) {
+
+        $response->setContent('Set by view handler');
+
+        return $response;
+    });
+
+View Handlers also receive the ``Request`` as its second argument,
+making them a good candidate for basic content negotiation::
+
+    $app->view(function (array $data, Request $request) use ($app) {
+
+        $acceptHeader = $request->headers->get('Accept');
+        $bestFormat = $app['negotiator']->getBestFormat($acceptHeader, array('json', 'xml'));
+
+        if ('json' === $bestFormat) {
+            return new JsonResponse($data);
+        }
+
+        if ('xml' === $bestFormat) {
+            return $app['serializer.xml']->renderResponse($data);
+        }
+
+        return $data;
+    });
+
+View Handlers will be examined in the order they are added to the application
+and Silex will use type hints to determine if a view handler should be used for
+the current result, continously using the return value of the last view handler
+as the input for the next::
+
+    $app->view(function (JsonResponse $response) use ($app) {
+        $response = $app['response_signer']->sign($response);
+
+        return $response;
+    });
+
+.. note::
+
+    You must ensure that Silex receives a ``Response`` or a string as the result of
+    the last view handler (or controller) to be run.
+
 Redirects
 ---------
 
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -542,49 +542,39 @@ early::
 View Handlers
 -------------
 
-View Handlers allow you to intercept a controller result and transform it before
-it gets returned to the kernel.
+View Handlers allow you to intercept a controller result that is not a
+``Response`` and transform it before it gets returned to the kernel.
 
 To register a view handler, pass a callable (or string that can be resolved to a
 callable) to the view method. The callable should accept some sort of result
 from the controller::
 
-    $app->view(function (Response $response) {
-
-        $response->setContent('Set by view handler');
-
-        return $response;
+    $app->view(function (array $controllerResult) use ($app) {
+        return $app->json($controllerResult);
     });
 
-View Handlers also receive the ``Request`` as its second argument,
+View Handlers also receive the ``Request`` as their second argument,
 making them a good candidate for basic content negotiation::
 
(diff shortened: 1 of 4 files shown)


src/Silex/Application.php
```diff
@@ -89,7 +89,7 @@ public function __construct(array $values = array())
 
         $this['dispatcher_class'] = 'Symfony\\Component\\EventDispatcher\\EventDispatcher';
         $this['dispatcher'] = $this->share(function () use ($app) {
-            /**
+            /*
```
stof: this should be kept as a docComment, as it is really one. the real fix would be to make it valid: ``@var EventDispatcherInterface $dispatcher``

src/Silex/ViewListenerWrapper.php
```diff
@@ -0,0 +1,88 @@
...
+class ViewListenerWrapper
+{
+    private $app;
+    private $callback;
+
+    /**
+     * Constructor.
+     *
+     * @param Application $app      An Application instance
+     * @param callable    $callback
```
stof: this may not be a callable yet here (it is a callable only after being resolved)

src/Silex/Application.php
```diff
@@ -405,6 +405,23 @@ public function error($callback, $priority = -8)
...
 
     /**
+     * Registers a view handler.
+     *
+     * View handlers are simple callables which take a controller result and the
+     * request as arguments, whenever a controller returns a value that is not
+     * an instance of Response. When this occurs, all suitable handlers will be
+     * called, until one returns a Response object.
+     *
+     * @param callable $callback View handler callback
```
stof: this is not (always) a callable here

src/Silex/Application.php
davedevelopment: Cheers @stof, learnt something new :)

tests/Silex/Tests/ApplicationTest.php
```diff
@@ -559,6 +559,128 @@ public function testRedirectDoesNotRaisePHPNoticesWhenMonologIsRegistered()
...
+        });
+
+        $response = $app->handle(Request::create('/foo'));
+
+        $this->assertEquals('Hello world', $response->getContent());
+    }
+
+    public function testViewListenerWithCallableTypeHint()
+    {
+        if (version_compare(phpversion(), '5.4.0', '<')) {
```
stof: use ``PHP_VERSION_ID`` here to compare the version (or ``@requires PHP 5.4`` in the docComment of the test, using the builtin PHPUnit feature)

src/Silex/Application.php
davedevelopment: Thanks @stof, fixed.

davedevelopment:
Also worth mentioning @fabpot [disliked the typehinting aspect](https://github.com/silexphp/Silex/pull/1062#issuecomment-70408305). Happy to take it out, just need a final confirmation.

fabpot:
I'm fine with the typehinting if people like it.

Crell:
The type hinting is quite useful, IMO. It makes it very easy to return multiple object types from controllers and have the right corresponding view listener "just work".

The branch looks good to me visually, although IMO it could stand some more comments, particularly in shouldRun(). :smile:  The fabbot disagrees though.

henrikbjorn:
:-1: on the typehinting magick. Using a guard isnt that difficult to understand or explain. The magic typehinting is hard to explain and understand.

anlutro:
I like typehinting magic, but Silex doesn't really do it much elsewhere, so for the sake of consistency I'd be inclined to leave it out.

mediamonks-robert:
I +1 the typehinting since it's basically the same as the error handler and it's not hard to understand at all.

Crell:
As Robert said, it's the same logic and largely the same code structure as exception handling already uses.  If it's not confusing for exceptions it shouldn't be confusing for view(), either.

davedevelopment:
I don't really mind about having the type hinting per se, but I do think in it's present state, this PR is consistent with the existing error handler behaviour, which is itself a kind of special case of view handling. 

I've added some docs, comments on those appreciated as well.

doc/usage.rst
```diff
@@ -539,6 +539,58 @@ early::
...
+-------------
+
+View Handlers allow you to intercept a controller result and transform it before
+it gets returned to the kernel.
+
+To register a view handler, pass a callable (or string that can be resolved to a
+callable) to the view method. The callable should accept some sort of result
+from the controller::
+
+    $app->view(function (Response $response) {
```
fabpot: I think this is unfortunate to use a Response here as ann example as this is probably not the common use case.

doc/usage.rst
```diff
@@ -539,6 +539,58 @@ early::
...
+
+        $response->setContent('Set by view handler');
+
+        return $response;
+    });
+
+View Handlers also receive the ``Request`` as its second argument,
```
fabpot: as their?

doc/usage.rst
```diff
@@ -539,6 +539,58 @@ early::
         return new Response(...);
     });
 
+View Handlers
+-------------
+
+View Handlers allow you to intercept a controller result and transform it before
+it gets returned to the kernel.
```
fabpot: You should mention that this is only when the controller result is not a response.

doc/usage.rst
GromNaN: A better example would be: 
```php
$app->view(function (array $controllerResult) use ($app) {
    return $app->json($controllerResult);
});
```

davedevelopment:
Replaced the ridiculously incorrect example, fixed the grammar and totally removed the last example, as I couldn't think of anything that wasn't a foobar type thing. Thanks @fabpot @GromNaN

fabpot:
Thank you @davedevelopment.

PR #1164, opened 2015-05-11
use legacy and refactorized symfony2 validator

Ma27:
with this pull request it is possible to use the legacy validator and the RecursiveValidator.

```php
// use legacy validator
$app->register(new ValidatorServiceProvider());
$validator = $app['validator']; // instance of Symfony\Component\Validator\Validator

// use RecursiveValidator
$app->register(new ValidatorServiceProvider(), array('use_refactorized_validator' => true));
$validator = $app['validator']; // instance of Symfony\Component\Validator\Validator\RecursiveValidator
```

### DIFF
--- a/src/Silex/Provider/ValidatorServiceProvider.php
+++ b/src/Silex/Provider/ValidatorServiceProvider.php
@@ -14,10 +14,13 @@ namespace Silex\Provider;
 use Silex\Application;
 use Silex\ServiceProviderInterface;
 use Silex\ConstraintValidatorFactory;
-use Symfony\Component\Validator\Validator;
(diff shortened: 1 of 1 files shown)


fabpot:
1.2 is a maintenance mode, so that cannot be accepted there. It should be on 1.3 where we already worked on supporting the newest versions of Symfony.

stof:
The ``use_refactorized_validator`` option is a bad idea anyway IMO. Silex should follow the same way than Symfony here: use the new implementation with BC layer (which should be the case automatically in 1.3 btw, as it uses the Symfony builder to configure the validator)

Issue #1152, opened 2015-04-18
Silex 2 and Symfony 2.7+ (with Twig 1.18.0)'s app.user


J7mbo:
Trying to use `{{ app.user }}` with the above versions says that it doesn't exist in the application. Is this because `twig-bridge` doesn't yet support it?

I have registered everything as expected, so I can still use `{{ app.security.token.user }}` but then I get some `E_DEPRECATED` notices. Again, switching to `{{ app.user }}` gives me `TWIG_ERROR_RUNTIME` saying that Method "user" for object does not exist.

What am I doing wrong?

J7mbo:
Addition: According to the [docs](http://symfony.com/doc/2.7/reference/twig_reference.html#app), it seems there should be this `GlobalVariables` class. I could create this myself, and I can only find reference to this within [FrameworkBundle](https://github.com/symfony/FrameworkBundle). Something missing in Silex / Twig?

J7mbo:
This was 5 days ago - am I doing something retarded?

GromNaN:
The documentation is related to Symfony fullstack, I agree that it's not obvious.

The `app` variable that is accessible in Twig refers to the Silex Application. So you have access to all the methods of this class and all the services registered. And by default there is no service or method named `user`.

Using the [`SecurityTrait`](http://silex.sensiolabs.org/doc/providers/security.html#traits) you can declare a method [`$app->user()`](https://github.com/silexphp/Silex/blob/master/src/Silex/Application/SecurityTrait.php#L32-L43) that returns the current user.

J7mbo:
As Silex previously had `app.security.token.user` available and it's deprecating accessing `app.security`, which is an instance of the deprecated `SecurityContext`, I'm presuming there needs to be a replacement using the new authorisation checker or token storage mechanisms. Otherwise, how is everyone using Silex going to access the user object, write their own or use the trait?

GromNaN:
It's `{{ security.token_storage.token.user }}`.

J7mbo:
Thanks @GromNaN... What is the security var an instance of?

GromNaN:
It's a `Symfony\Component\Security\Core\SecurityContext`.
https://github.com/silexphp/Silex/blob/master/src/Silex/Provider/SecurityServiceProvider.php#L95-L97

stof:
use ``app.token_storage`` to access the TokenStorage (given that ``app`` is your Silex Application)

stof:
[...]

PR #1165, opened 2015-05-14
Easier user access

fabpot:
In Silex 1.3, it's "harder" to get the user from the security layer as the `security` service has been deprecated. The code is already available in the security trait, but if you are not using it, you need to use a somewhat "complex" expression in templates to access it (see #1152)

This PR moves the code from the trait to the security service to make it available to everyone.

One question though: can we deprecate the trait `user()` function? It is not useful anymore after this PR.

### DIFF
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -101,6 +101,18 @@ class SecurityServiceProvider implements ServiceProviderInterface
             });
         }
 
+        $app['user'] = function ($app) {
+            if (null === $token = $app['security.token_storage']->getToken()) {
+                return;
+            }
+
+            if (!is_object($user = $token->getUser())) {
+                return;
+            }
+
+            return $user;
+        };
+
         $app['security.authentication_manager'] = $app->share(function ($app) {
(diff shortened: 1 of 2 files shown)


J7mbo:
I thought using `$app['security.token_storage']` was deprecated, and would throw out some `E_DEPRECATED`s... or am I mistaken?

Kilte:
@J7mbo 
>or am I mistaken?

This

`security_context` was deprecated.

GromNaN:
:+1: For deprecating the trait and removing it from Silex 2.0

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -101,6 +101,18 @@ public function register(Application $app)
             });
         }
 
+        $app['user'] = function ($app) {
```
stof: this should not be a shared method. Otherwise you will get the wrong value if you access it first before the firewall gets call and then a second time after (the second time will get the cached ``null`` value)

fabpot: It's not a shared method, we are in 1.3 here.

stof: ah sorry, I missed it

PR #1167, opened 2015-05-19
Update fastcgi_pass socket location

victorhaggqvist:
This makes it correspond with ubuntu 14.04 defaults

### DIFF
--- a/doc/web_servers.rst
+++ b/doc/web_servers.rst
@@ -61,7 +61,11 @@ resources to ``index.php``:
         }
 
         location @site {
-            fastcgi_pass   unix:/var/run/php-fpm/www.sock;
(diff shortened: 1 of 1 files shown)


fabpot:
But then, what about other versions? I think we should keep both paths and add a comment about which distrib is using what.

victorhaggqvist:
That is likely a good idea. From what I could catch online the one present atm,  `unix:/var/run/php-fpm/www.sock`, would be the CentOS default, right?

victorhaggqvist:
Maybe something like that. Don't know which one would be better as the default uncommented, but did set it as ubuntu now.

fabpot:
Thank you @victorhaggqvist.

PR #1172, opened 2015-06-08
Change that if the app locale is empty use the one from the locale listener

rsantellan:
Setting the locale with $app['translator']->setLocale($lang); was always override by the listener on the onKernelRequest method:

$this->app['locale'] = $event->getRequest()->getLocale();

the $event->getRequest()->getLocale() was NULL so the default was given setting that to 'en' 

Changing that check first is a locale is set fix that problem.

### DIFF
--- a/src/Silex/EventListener/LocaleListener.php
+++ b/src/Silex/EventListener/LocaleListener.php
@@ -36,7 +36,9 @@ class LocaleListener extends BaseLocaleListener
     public function onKernelRequest(GetResponseEvent $event)
     {
         parent::onKernelRequest($event);
-
-        $this->app['locale'] = $event->getRequest()->getLocale();
+        if($this->app['locale'] === null)
+        {
+          $this->app['locale'] = $event->getRequest()->getLocale();
+        }
     }
 }
--- a/src/Silex/EventListener/LocaleListener.php
+++ b/src/Silex/EventListener/LocaleListener.php
@@ -36,8 +36,7 @@ class LocaleListener extends BaseLocaleListener
     public function onKernelRequest(GetResponseEvent $event)
     {
         parent::onKernelRequest($event);
-        if($this->app['locale'] === null)
-        {
+        if($this->app['locale'] === null){
           $this->app['locale'] = $event->getRequest()->getLocale();
         }
     }


src/Silex/EventListener/LocaleListener.php
```diff
@@ -36,7 +36,9 @@ public function __construct(Application $app, RequestContextAwareInterface $rout
     public function onKernelRequest(GetResponseEvent $event)
     {
         parent::onKernelRequest($event);
-
-        $this->app['locale'] = $event->getRequest()->getLocale();
+        if($this->app['locale'] === null)
+        {
```
davedevelopment: This opening bracket needs to be at the end of the line above

stof:
this look wrong to me. Silex is documenting that ``$app['locale']`` gets set to the current locale, and it is not the case anymore after your change

rsantellan:
The problem is that **$event->getRequest()->getLocale();** always is returning the default language for me defined on: **Symfony\Component\HttpFoundation\Request** because **$this->locale** is NULL
```php
    /**
     * @var string
     */
    protected $defaultLocale = 'en';
```
That make impossible to do change the language using something like this on my **app.php**
```php
$lang = "en";
if ($app['session']->get('current_language')) {
  $lang = $app['session']->get('current_language');
} else{
  $app['session']->set('current_language', $lang);
}
$app['translator']->setLocale($lang);  
```
Changing that make the same behaviour I was having with a previous release. This are the versions of my composer.json of a old project.
```
Name: __root__
Version: 9999999-dev
Licenses: none
Dependencies:
    
  doctrine/annotations       v1.2.4    MIT           
  doctrine/cache             v1.4.1    MIT           
  doctrine/collections       v1.3.0    MIT           
  doctrine/common            v2.5.0    MIT           
  doctrine/dbal              v2.5.1    MIT           
  doctrine/inflector         v1.0.1    MIT           
  doctrine/lexer             v1.0.1    MIT           
  jasongrimes/paginator      1.0.1     MIT           
  monolog/monolog            1.13.1    MIT           
  pimple/pimple              v1.1.1    MIT           
  psr/log                    1.0.0     MIT           
  silex/silex                v1.2.4    MIT           
  symfony/config             v2.6.6    MIT           
  symfony/debug              v2.6.6    MIT           
  symfony/event-dispatcher   v2.6.6    MIT           
  symfony/filesystem         v2.6.6    MIT           
  symfony/form               v2.6.6    MIT           
  symfony/http-foundation    v2.6.6    MIT           
  symfony/http-kernel        v2.6.6    MIT           
  symfony/intl               v2.6.6    MIT           
  symfony/options-resolver   v2.6.6    MIT           
  symfony/property-access    v2.6.6    MIT           
  symfony/routing            v2.6.6    MIT           
  symfony/security           v2.6.6    MIT           
  symfony/translation        v2.6.6    MIT           
  symfony/twig-bridge        v2.6.6    MIT           
  symfony/validator          v2.6.6    MIT           
  symfony/yaml               v2.6.6    MIT           
  twig/twig                  v1.18.1   BSD-3-Clause 
```

rsantellan:
Any news for this? or the correct way to change the locale?

Regards.

CarsonF:
@rsantellan Why not create a "SessionLocaleListener" that subscribes to `kernel.request` with a priority after `SessionListener/SessionServiceProvider` (128) and before `LocaleListener` (16) that sets the request locale from the session? That in turn would set `$app['locale']` and translator locale.

It doesn't even have to be a listener class, i.e.:
```php
$app->before(function(Request $request) {
    $session = $request->getSession();
    if ($session->has('current_language')) {
        $request->setLocale($session->get('current_language'));
    }
}, 17);
```

rsantellan:
@CarsonF the solution that you give works really good. It should be on the documentation for easy access in the translation part.

Thank very much.

PR #1174, opened 2015-06-11
Added factory for RouteCollection, makes subclassing of RouteCollection possible.

lschricke:
The aim of this PR is to allow developpers to subclass `Symfony\Component\Routing\RouteCollection` to change some of its behavior.

In most of my Silex apps, I figured I don't want the Symfony's `RouteCollection` regular behavior where if you define two routes with the same name (bad copy-paste for instance), the first route added simply gets overriden by the second. Instead, I would like to be able to throw an Exception if that happens. So I've written [a tiny subclass of RouteCollection](https://github.com/lschricke/symfony-strict-route-collection), but to be able to use it (through [a service provider like this](https://github.com/lschricke/silex-strict-route-collection-service-provider) for instance), I need to be able to replace `RouteCollection` by my subclass. This PR makes it possible.

Thanks! :)


fabpot:
Not sure if we want to support this use case. Being able to override the route collection means that we can potentially have very difficult to spot bugs. The fact that you can override a route is indeed a feature that makes it easy to override a third-party module.

lschricke:
I understand that this could lead to difficult problems if not used carefully. But I tend to assume that if you override any default class (or behavior in general), you propably know (at least to some extent) what you are doing and that you have to be more careful with your code. Do you think this change is that risky?

There may be a better or different way to do that (throwing an exception instead of overriding already added routes), do you know if some other developers also ran into this issue and found other ways to handle it? I mean, Silex being a micro-framework, I think in many cases you don't want the routes you defined to be overridden at all. Am I wrong about that?

fabpot:
It's not an issue but a feature that people rely on. Anyway, the changes are not backward compatible and could only be considered for the master branch.

lschricke:
I'm sorry, I misused the word 'issue'. I meant use case.

lschricke:
You're right, I skipped that this would break BC. I'm closing this PR and will resubmit it to the master branch.

PR #1175, opened 2015-06-12
Added factory for RouteCollection, makes subclassing of RouteCollection possible.

lschricke:
Note: replaces #1174, which was targeting a wrong branch (and where I messed a bit with the commits ;)).

--

The aim of this PR is to allow developpers to subclass `Symfony\Component\Routing\RouteCollection` to change some of its behavior.

In most of my Silex apps, I figured I don't want the Symfony's `RouteCollection` regular behavior where if you define two routes with the same name (bad copy-paste for instance), the first route added simply gets overriden by the second. Instead, I would like to be able to throw an Exception if that happens. So I've written [a tiny subclass of RouteCollection](https://github.com/lschricke/symfony-strict-route-collection), but to be able to use it (through [a service provider like this](https://github.com/lschricke/silex-strict-route-collection-service-provider) for instance (needs to be updated for Silex 2)), I need to be able to replace `RouteCollection` by my subclass. This PR makes it possible.

Thanks! :)

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -64,8 +64,11 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
 
         $app = $this;
 
-        $this['routes'] = function () {
+        $this['routes_factory'] = $this->factory(function () {
             return new RouteCollection();
+        });
+        $this['routes'] = function () use ($app) {
+            return $app['routes_factory'];
         };
 
         $this['controllers'] = function () use ($app) {
@@ -73,7 +76,7 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
         };
 
         $this['controllers_factory'] = $this->factory(function () use ($app) {
-            return new ControllerCollection($app['route_factory']);
+            return new ControllerCollection($app['route_factory'], $app['routes_factory']);
         });
 
         $this['route_class'] = 'Silex\\Route';
--- a/src/Silex/ControllerCollection.php
+++ b/src/Silex/ControllerCollection.php
@@ -42,10 +42,12 @@ class ControllerCollection
     protected $defaultRoute;
     protected $defaultController;
     protected $prefix;
+    protected $routesFactory;
 
-    public function __construct(Route $defaultRoute)
+    public function __construct(Route $defaultRoute, $routesFactory = null)
     {
         $this->defaultRoute = $defaultRoute;
+        $this->routesFactory = $routesFactory;
         $this->defaultController = function (Request $request) {
             throw new \LogicException(sprintf('The "%s" route must have code to run when it matches.', $request->attributes->get('_route')));
         };
@@ -186,7 +188,11 @@ class ControllerCollection
      */
     public function flush($prefix = '')
     {
-        $routes = new RouteCollection();
+        if (null === $this->routesFactory) {
+            $routes = new RouteCollection();
+        } else {
+            $routes = $this->routesFactory;
+        }
 
         foreach ($this->controllers as $controller) {
             if ($controller instanceof Controller) {
--- a/tests/Silex/Tests/ApplicationTest.php
+++ b/tests/Silex/Tests/ApplicationTest.php
@@ -23,6 +23,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\Routing\RouteCollection;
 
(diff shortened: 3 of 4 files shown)


src/Silex/Application.php
```diff
@@ -64,8 +63,12 @@ public function __construct(array $values = array())
 
         $app = $this;
 
-        $this['routes'] = function () {
-            return new RouteCollection();
+        $this['routes_class'] = 'Symfony\\Component\\Routing\\RouteCollection';
```
fabpot: I would remove this as overriding the `routes_factory` would do the same. So, it's not really a good extension point.

src/Silex/Application.php
```diff
@@ -64,8 +63,12 @@ public function __construct(array $values = array())
...
+        $this['routes_factory'] = $this->factory(function () use ($app) {
```
fabpot: I don't see the need for this one as `$this['routes']` looks enough.

src/Silex/Application.php
```diff
@@ -73,7 +76,7 @@ public function __construct(array $values = array())
         };
 
         $this['controllers_factory'] = $this->factory(function () use ($app) {
-            return new ControllerCollection($app['route_factory']);
+            return new ControllerCollection($app, $app['route_factory']);
```
fabpot: I would make the new argument second instead of prepending it to avoid an unneeded BC break. I would also inject the routes factory instead of the whole app.

src/Silex/ControllerCollection.php
```diff
@@ -38,13 +37,15 @@
...
 class ControllerCollection
 {
+    protected $app;
     protected $controllers = array();
     protected $defaultRoute;
     protected $defaultController;
     protected $prefix;
 
-    public function __construct(Route $defaultRoute)
+    public function __construct(Application $app, Route $defaultRoute)
```
fabpot: So, the routes factory would be the second argument and it can even be optional. If not provided, just create a route collection like before.

tests/Silex/Tests/ApplicationTest.php
```diff
@@ -441,7 +441,7 @@ public function testRegisterShouldReturnSelf()
     public function testMountShouldReturnSelf()
     {
         $app = new Application();
-        $mounted = new ControllerCollection(new Route());
+        $mounted = new ControllerCollection($app, new Route());
```
fabpot: As the new argument is now optional, the tests can be keep like before, but new tests should be added to cover the new feature.

src/Silex/Application.php
stof: a factory is needed if ControllerCollection reuses it.

However @lschricke, this PR won't solve things for you if a ControllerCollection defines the same route name than another one (or that the main route collection)

lschricke: Thanks for considering this PR and for the feedback! I'll look into it and update the PR soon.

lschricke:
I updated the code and added some tests to cover this new feature as you suggested @fabpot. Do you think the tests provide enough coverage?

@stof, this PR does not solve my use case directly, but using it to have `Silex` use a subclass of Symfony's `RouteCollection` like [this one](https://github.com/lschricke/symfony-strict-route-collection) seems to work in my tests. Am I missing something?

src/Silex/Application.php
```diff
@@ -29,12 +29,12 @@
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpFoundation\JsonResponse;
-use Symfony\Component\Routing\RouteCollection;
 use Silex\Api\BootableProviderInterface;
 use Silex\Api\EventListenerProviderInterface;
 use Silex\Api\ControllerProviderInterface;
 use Silex\Provider\RoutingServiceProvider;
 use Silex\Provider\KernelServiceProvider;
+use Symfony\Component\Routing\RouteCollection;
```
fabpot: This should be reverted back to its original place.

src/Silex/ControllerCollection.php
```diff
@@ -11,8 +11,8 @@
 
 namespace Silex;
 
-use Symfony\Component\Routing\RouteCollection;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\RouteCollection;
```
fabpot: should be moved back as before.

lschricke:
@fabpot PR updated.

lschricke:
I squashed my commits and rebase the PR.

fabpot:
Thank you @lschricke.

lschricke:
And thank you @fabpot and all the contributors for this awesome tool! :)

PR #1177, opened 2015-06-12
Removing the default array to allow for $app config to be used

cgsmith:
This would allow $app to have the config options passed to swiftmailer's provider.


stof:
This PR is wrong. All Silex providers are expecting you to configure them **after** registering them. So setting a default values for the needed parameter is the expected way.

cgsmith:
I did not know that.  I'm curious then, why would it do an array replace on the $options?

cgsmith:
@stof how would i do something similar to how the doctrine service provider instantiates?

If i can do something similar i can always submit a PR to the docs as this was unclear which is why i ended up down this original path.  Excuse my mistake if i made one.

cgsmith:
I suppose I have had a rtfm moment.... 

https://github.com/silexphp/Silex/blob/1.3/doc/providers.rst

PR #934, opened 2014-04-01
Allow setting of Session FlashBag and AttributeBag.

robomc:
This change allows you to set the FlashBag and AttributeBag for the Symfony Session when registering the SessionServiceProvier:

    $app->register(new Silex\Provider\SessionServiceProvider(), array(
        'session.storage.options' => array(
          'name' => 'mycookie'
        ),
        'session.flashbag' => new Symfony\Component\HttpFoundation\Session\Flash\AutoExpireFlashBag
      )
    );

Allows options equivalent to Symfony's `session.flashbag.class` and `session.attribute_bag.class` options.

PR #1178, opened 2015-06-12
Session refactoring

fabpot:
replaces #934 by making possible to override the default bags, but more importantly, it removes the storage definition from the inside of the session one to the outside.


SpacePossum:
:+1:  thanks!

PR #1183, opened 2015-06-18
create logdir if not exists

dominikzogg:
Would be cool , if the provider guaranteed, that the log folder exists, where the logfile will be storred in.

### DIFF
--- a/src/Silex/Provider/MonologServiceProvider.php
+++ b/src/Silex/Provider/MonologServiceProvider.php
@@ -55,6 +55,14 @@ class MonologServiceProvider implements ServiceProviderInterface
 
         $app['monolog.handler'] = function () use ($app) {
             $level = MonologServiceProvider::translateLevel($app['monolog.level']);
+            if (is_string($app['monolog.logfile']) && false === strpos($app['monolog.logfile'], '://')) {
+                $logdir = dirname($app['monolog.logfile']);
+                if (!is_dir($logdir)) {
+                    if (!@mkdir($logdir, 0755, true)) {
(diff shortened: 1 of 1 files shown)


dominikzogg:
feedback from irc #silex-php

13:13:18 - theahindle: There are much more knowledgeable people than me, but I personally think it shouldn't change the file system in the background - it should be explicit
13:13:25 - steffkes: welcome in the nice world of php ..
13:13:43 - theahindle: Like setting the handler to Monolog\LoggerMakeDirOnMyBehalfPlease or setting $app['monolog.mkdir'] = true;
13:13:53 - steffkes: yeah i like that more as well theahindle.

dominikzogg:
add an option is a valid solution for me, @fabpot what do you think?

henrikbjorn:
In any case it should not be done in the definition of the service.

dominikzogg:
@henrikbjorn Well what is your suggestion, close this PR and create one within monolog or handle it in my project(s)?

henrikbjorn:
Yes. Monolog should do this inside the handler that writes to the filesystem. Until then, you should handle it in your projects.

On the code side of things. You are adding complexity through 3 new indentation levels and many more code paths.

fabpot:
This is clearly the responsibility of Monolog to do that (if they want to, and they should :))

Regarding the option, that would not make sense. Asking someone to set an option to create a directory seems like the same amount of work than creating the directory directly. Closing this PR.

dominikzogg:
@fabpot https://github.com/Seldaek/monolog/pull/584 should be solved

fabpot:
@dominikzogg thanks for the heads up, that's a good news!

PR #1186, opened 2015-06-25
Update session_storage.rst: Add missing lifetime column

andytruong:


fabpot:
Thank you @andytruong.

PR #1190, opened 2015-07-08
Allow custom logout handlers when using custom authentication

mivanov93:
When using a custom authentication factory:
As of now, one can only select a Logout Success Handler(which is erroneously called logout handler, but I left the old name for compatibility's sake) which is called after the Logout handler was called and the logout was successful.
But custom authentication may need a custom Logout handler.(one may not want to use sessions at all, but custom cookies/etc.)
I've called it logout_clear_data_handler, since it clears data. And by default it should be SessionLogoutHandler();(the prototype for it).

### DIFF
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -488,6 +488,12 @@ class SecurityServiceProvider implements ServiceProviderInterface
                 );
             });
         });
+        
+        $app['security.authentication.logout_clear_data_handler._proto'] = $app->protect(function ($name, $options) use ($app) {
+            return $app->share(function () use ($name, $options, $app) {
+                return new SessionLogoutHandler();
+            });
+        });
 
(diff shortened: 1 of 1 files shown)


mivanov93:
The automatic check has incorrectly failed:
fabbot.io — Some changes should be done to comply with our standards. 

I've checked the patch it generated:
It says to remove a newline and then add it again - which obviously is nonsensical statement.

mathroc:
not exactly, it told you to remove the tabs and spaces of those empty lines (if you select the lines you can see there is spaces in your lines and not in the ones added)

mivanov93:
Done, thank you for noticing :+1:

fabpot:
The logout handler does not need not need to be configurable, but it should be indeed possible to disable it like it's done in Symfony. So, closing this one in favor of #916 which is more in line with Symfony full-stack behavior.

mivanov93:
Wouldn't it be better to do both? I The disabling is important too I think, but choosing a custom logout handler seems a matter of importance. For example I'd rather use custom cookies for my sessions.(I have my reasons).

PR #1192, opened 2015-07-14
Added a logout handler and its prototype for clearing data

mivanov93:
Already committed in the 1.3 branch and waiting for a merge, since it works fine for the 2.0~ branch, pushing it here as well.(only change being $app->share(f) to f, since it's Pimple 3 and not 1.(the reason for this additional commit and not just in the 1.3 branch)


fabpot:
Closing as branches are merged from the oldest one the the newest ones, so if the other PR is merged, it will be merged into master as well.

PR #1193, opened 2015-07-16
Fix PHPDoc return type of WebTestCase

SpacePossum:
The test expects an `Application` instance but `HttpKernel` was hinted.

### DIFF
--- a/src/Silex/WebTestCase.php
+++ b/src/Silex/WebTestCase.php
@@ -12,7 +12,6 @@
 namespace Silex;
 
 use Symfony\Component\HttpKernel\Client;
-use Symfony\Component\HttpKernel\HttpKernel;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Silex/WebTestCase.php
```diff
@@ -42,14 +41,14 @@ public function setUp()
     /**
      * Creates the application.
      *
-     * @return HttpKernel
+     * @return Application
```
GromNaN: You can return any `HttpKernel` (like the `HttpCache` or any StackPHP middleware).

SpacePossum: ah thanks, so it should be `HttpKernelInterface`? (since Application does not extend `HttpKernel`)

GromNaN: Yes.

PR #1194, opened 2015-07-16
PHPDocs Application -> HttpKernelInterface

SpacePossum:
replacement of https://github.com/silexphp/Silex/pull/1193
(could not reopen that one because of a forced push)

### DIFF
--- a/src/Silex/WebTestCase.php
+++ b/src/Silex/WebTestCase.php
@@ -12,7 +12,7 @@
 namespace Silex;
 
 use Symfony\Component\HttpKernel\Client;
-use Symfony\Component\HttpKernel\HttpKernel;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
 
 /**
  * WebTestCase is the base class for functional tests.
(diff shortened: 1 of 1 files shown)


src/Silex/ServiceProviderInterface.php
```diff
@@ -23,6 +23,8 @@
      *
      * This method should only be used to configure services and parameters.
      * It should not get services.
+     *
+     * @param Application $app
```
fabpot: Can be removed as all the info are already part of the function definition.

src/Silex/ServiceProviderInterface.php
```diff
@@ -32,6 +34,8 @@ public function register(Application $app);
      * This method is called after all services are registered
      * and should be used for "dynamic" configuration (whenever
      * a service must be requested).
+     *
+     * @param Application $app
```
fabpot: same here

SpacePossum: check, thanks and done :)

fabpot:
Thank you @SpacePossum.

Issue #1196, opened 2015-07-19
SecurityServiceProvider: custom authentication listener is never registered


SanderKnape:
The SecurityServiceProvider never registers a custom authentication listener, making it impossible to add a custom authentication provider. Looking at [line 245](https://github.com/silexphp/Silex/blob/1.3/src/Silex/Provider/SecurityServiceProvider.php#L245), the required components are retrieved (provider, listener, entry point). The listener is then added to the `$factories` array on [line 251](https://github.com/silexphp/Silex/blob/1.3/src/Silex/Provider/SecurityServiceProvider.php#L251). However, in the following [for loop](https://github.com/silexphp/Silex/blob/1.3/src/Silex/Provider/SecurityServiceProvider.php#L255-L259), the hard-coded array `$positions` defined earlier is used to gather all listeners. This array does not contain the custom listener, thus it is never registered.

The easiest solution would probably be to add the following lines somewhere around line 251:

    if(!in_array($position, $positions)) {
      $positions[] = $position;
    }

fabpot:
The list of positions is fixed and non-extandable. So, your factory must return one of the available ones.

mivanov93:
You need to create a factory, best way would be to make a provider and put the factory function in the register function. Basically create $app\['security.authentication_listener.factory.'.$name\](where name is the name you're going to call your authentication, not the name of the firewall) as function($app['example']=$app->protect(function(){});) and it must return an array of a provider, listener,etc. which will be used by the SecurityServiceProvider.

PR #1198, opened 2015-07-22
Fix custom authentication providers not being registered.

SanderKnape:
Fix for Issue #1196.


SanderKnape:
Ugh. The check keeps failing because composer can't [update its dependencies](https://travis-ci.org/silexphp/Silex/jobs/72148198) for the php7 build. Anyone has any tips?

PR #1200, opened 2015-07-25
Allowing the use of both objects and arrays for the register()

mivanov93:
Sometimes we may like to use objects for lazy loading and whatsoever and since PHP has the Traversable interface, that should be perfectly fine.

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -162,11 +162,11 @@ class Application extends \Pimple implements HttpKernelInterface, TerminableInte
      * Registers a service provider.
      *
      * @param ServiceProviderInterface $provider A ServiceProviderInterface instance
-     * @param array                    $values   An array of values that customizes the provider
+     * @param array|Traversible        $values   An array or a Traversible of values that customizes the provider
      *
      * @return Application
      */
-    public function register(ServiceProviderInterface $provider, array $values = array())
+    public function register(ServiceProviderInterface $provider, $values = array())
     {
         $this->providers[] = $provider;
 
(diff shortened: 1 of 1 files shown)


src/Silex/Application.php
```diff
@@ -162,11 +162,11 @@ public function __construct(array $values = array())
      * Registers a service provider.
      *
      * @param ServiceProviderInterface $provider A ServiceProviderInterface instance
-     * @param array                    $values   An array of values that customizes the provider
+     * @param array|Traversible        $values   An array or a Traversible of values that customizes the provider
```
mathroc: the name is [`Traversable`](http://php.net/manual/en/class.traversable.php)

src/Silex/Application.php
```diff
@@ -162,11 +162,11 @@ public function __construct(array $values = array())
...
      *
      * @return Application
      */
-    public function register(ServiceProviderInterface $provider, array $values = array())
+    public function register(ServiceProviderInterface $provider, $values = array())
```
mathroc: instead of just removing `array` you should typehint with `Traversable`

mivanov93:
Thanks, fixed the typo.

fabpot:
Can you rebase this PR (that will remove the merge commit that you have here)?

fabpot:
As the method signature changed, it can only be merged in master.

fabpot:
Thank you @mivanov93.

fabpot:
Reverted as this breaks the Pimple interface. So, we would need to also break BC on Pimple to make this happen, which I think is not worth it.

PR #1204, opened 2015-07-30
Add form type support in FormTrait for flexibility

lokhman:
Set parameter to the end for backwards compatibility, ideally make it first in the future release.

### DIFF
--- a/src/Silex/Application/FormTrait.php
+++ b/src/Silex/Application/FormTrait.php
@@ -25,11 +25,12 @@ trait FormTrait
      *
      * @param mixed $data    The initial data for the form
      * @param array $options Options for the form
+     * @param mixed $type    Type of the form
      *
      * @return FormBuilder
      */
-    public function form($data = null, array $options = array())
+    public function form($data = null, array $options = array(), $type = 'form')
     {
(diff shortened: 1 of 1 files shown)


fabpot:
That can only be merged in the master branch.

lokhman:
Yes, that's my bad. Shall I close this request and open the new one for master commit then?

fabpot:
@lokhman No, I can merge this PR directly into master.

lokhman:
@fabpot Brilliant, thanks! Will it be possible to make `$type` argument coming first at some point in the future release? I personally find it rather important when extending `form.types` with own forms.

src/Silex/Application/FormTrait.php
```diff
@@ -25,11 +25,12 @@
      *
      * @param mixed $data    The initial data for the form
      * @param array $options Options for the form
+     * @param mixed $type    Type of the form
```
stof: why ``mixed`` ? The only supported types are ``FormTypeInterface`` and ``string``

lokhman: True. Fixed it now to more correct.

fabpot:
merged into master, thanks.

PR #1206, opened 2015-08-03
removed flush() argument

fabpot:
I propose to remove the `flush()` prefix argument as I don't see any use case. I think it was there because the method is recursive, but that's leaking an implementation detail. Passing a prefix was not documented.


stof:
:+1:

PR #1207, opened 2015-08-03
changed auto-generated route names when there is a conflict

fabpot:


src/Silex/ControllerCollection.php
```diff
@@ -208,8 +208,9 @@ private function doFlush($prefix, RouteCollection $routes)
                 $controller->getRoute()->setPath($prefix.$controller->getRoute()->getPath());
                 if (!$name = $controller->getRouteName()) {
                     $name = $controller->generateRouteName('');
+                    $i = 0;
                     while ($routes->get($name)) {
-                        $name .= '_';
+                        $name = $controller->generateRouteName('').'_'.++$i;
```
stof: I suggest avoid to redo the generation multiple times. Just keep the original generated name in a separate variable

Issue #1044, opened 2014-10-24
The "host" authentication entry is not registered.


jazzzpor:
It seems the "host" parameter is not yet implemented in the Security Service Provider. 
In Symfony it has been available since version 2.4: http://symfony.com/doc/current/cookbook/security/firewall_restriction.html

Is there a way around this? I need the firewall to initialise only on a specific subdomain.

PR #1208, opened 2015-08-03
updated the docs to show how to restrict a firewall by more than just the URL

### LINKED ISSUE #1171: Firewall specific HTTP methods
I've found out that in the Symfony Security component you can restrict the firewall to be triggered upon a specific request ([documentation](http://symfony.com/doc/current/cookbook/security/firewall_restriction.html#restricting-by-http-methods)). 

In my case, for a certain endpoint of my API, I'd like the firewall to be triggered on a POST request but not on a GET request. Based on the documentation above I was planning to use something like the following when defining my security rules for Silex in a yaml file:

```yaml
endpoint:
  pattern: ^/myendpoint/$
  methods: [GET]
  security: false
default:
  pattern: ^.*$
  oauth2: true
```

But, checking on how the SecurityServiceProvider will handle that config, I've realized that the `methods` entry will be ignored: 
- No check for `methods` key ([code](https://github.com/silexphp/Silex/blob/1.3/src/Silex/Provider/SecurityServiceProvider.php#L199))
- Request matcher built only with one param [code](https://github.com/silexphp/Silex/blob/1.3/src/Silex/Provider/SecurityServiceProvider.php#L287) when it could accept more [code](https://github.com/symfony/HttpFoundation/blob/master/RequestMatcher.php#L61)

Is this done for a particular reason or it's an oversight in Silex?
Could this behaviour be achieved in some other way?

fabpot:
closes #1171 and #1044

### DIFF
--- a/doc/providers/security.rst
+++ b/doc/providers/security.rst
@@ -136,10 +136,24 @@ under ``/admin/``::
         ),
     );
 
-The ``pattern`` is a regular expression (it can also be a `RequestMatcher
+The ``pattern`` is a regular expression on the URL path; the ``http`` setting
+tells the security layer to use HTTP basic authentication and the ``users``
+entry defines valid users.
+
+If you want to restrict the firewall by more than the URL pattern (like the
+HTTP method, the client IP, the hostname, or any Request attributes), use an
+instance of a `RequestMatcher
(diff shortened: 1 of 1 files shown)


cuevae:
+1 this is awesome!

PR #1211, opened 2015-08-04
Allow to change variable name with converter

mathroc:
sometimes the name of a route parameter does not fit exactly for the name of the converted object, this allows to optionally set a new name for a converted variable

usage:

```php
$app->get('/user/:username', function(UserInterface $user) {
    // ...
})->convert('username', 'provider.user: loadUserByUsername', 'user');
```

### DIFF
--- a/src/Silex/EventListener/ConverterListener.php
+++ b/src/Silex/EventListener/ConverterListener.php
@@ -49,10 +49,12 @@ class ConverterListener implements EventSubscriberInterface
         $request = $event->getRequest();
         $route = $this->routes->get($request->attributes->get('_route'));
         if ($route && $converters = $route->getOption('_converters')) {
-            foreach ($converters as $name => $callback) {
+            foreach ($converters as $name => list($callback, $newName)) {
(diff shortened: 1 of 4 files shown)


fabpot:
The first argument of the converter is actually the variable of the controller, not the one in the route. And indeed, there such an example in the documentation:

```php
$callback = function ($post, Request $request) {
    return new Post($request->attributes->get('slug'));
};

$app->get('/blog/{id}/{slug}', function (Post $post) {
    // ...
})->convert('post', $callback);
```

mathroc:
oh, I didn't saw that the Request was given to the converter. still, it is a bit cumbersome to use the Request in converter (especially if those are services like the provider.user in my last exemple) 
currently, the first parameter of the converter is null if the controller parameter name does not exists in the route. would you be open to another approach where the third parameter can overwrite which route parameter is given to the converter ?

which would be used like that :

```php
$app->get('/user/:username', function(UserInterface $user) {
    // ...
})->convert('user', 'provider.user: loadUserByUsername', 'username');
```

Issue #1176, opened 2015-06-12
SwiftmailerServiceProvider overwrites default $app['swiftmailer.options'] on instantiation


cgsmith:
Swiftmailer's register() does not allow it to set the options.  It just overrides the app's options that may have been set previously.  You can see this on line 26 of the SwiftmailerServiceProvider.php file

GromNaN:
I don't have any issue with swiftmailer options. Can you paste a code sample ?

cgsmith:
I am using @igorw config provider.  I have examples but am headed out of the country shortly.  I can provide an example when back.

But if you extend the slim application and overload the constructor you might be able to replicate (i call the parent right after that).

I overload the constructor to have my class register the providers i need.  First Igor's config, then swiftmailer.

I found that $app['swiftmailer.options'] = array (); was causing the assignment issues.

Sorry for any readability issues here.  I am on the phone.

cgsmith:
Sorry.  Apparently an rtfm moment

Sorry to waste time!


https://github.com/silexphp/Silex/blob/1.3/doc/providers.rst

kamilsk:
please show code examples

cgsmith:
I will by the end of today.  Found out where the inconsistency was.

cgsmith:
Hey @kamilsk - I found out this issue a bit later playing around with Igor's config service provider.  If you already have `$app['swiftmailer.options']`, like you would using the Doctrine service provider, it will ignore them.

With [this commit](https://github.com/cgsmith/Silex/commit/70c056f5d1b998d76d10b5833c38562377df21ab) it will be able to allow it.

Test the issue with the following code:

```php
// The issue lies here

$app['swiftmailer.options'] = array('host'=>'your-host-name-here');
$app->register(new Silex\Provider\SwiftmailerServiceProvider());
var_dump($app['swiftmailer.options']); // This will have 'host' as 'localhost' as that is the default
```

Looking at the [Doctrine service provider](https://github.com/silexphp/Silex/blob/1.3/src/Silex/Provider/DoctrineServiceProvider.php#L48), you can see that they allow defaults to be set before the provider is instantiated.

**Note**: I don't think there would be any BC breaks due to the fact that most people would of had to configure the provider after loading it.

If all looks good I will submit the PR.

kamilsk:
Your example is not a "true way", IMO. If you use that provider as specified in the documentation, then everything is working fine.

See
[...]

PR #1214, opened 2015-08-13
Creating consistent provider for SwiftMailer

cgsmith:
This is a pull request for Issue #1176 

It addresses some provider inconsistencies and also provides a test according to how I think the provider would function.  This allows the programmer to utilize the paradigm they choose and fits better with other Config Service Providers out there.

### DIFF
--- a/src/Silex/Provider/SwiftmailerServiceProvider.php
+++ b/src/Silex/Provider/SwiftmailerServiceProvider.php
@@ -23,7 +23,7 @@ class SwiftmailerServiceProvider implements ServiceProviderInterface
 {
     public function register(Application $app)
     {
-        $app['swiftmailer.options'] = array();
+        $app['swiftmailer.options'] = isset($app['swiftmailer.options']) ? $app['swiftmailer.options'] : array();
         $app['swiftmailer.use_spool'] = true;
 
         $app['mailer.initialized'] = false;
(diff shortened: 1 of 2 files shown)


stof:
-1 for this. It is not consistent with the Silex conventions. All providers are defining the default values of their parameters directly.

The doctrine provider is a bit special because it allows to configure multiple connection, and ``db.default_options`` are the default options for **all** connections while the optional parameter ``db.options`` (which does not have any default value) is a shortcut to define options for the main connection to simplify the case of a single connection.
and the ``db.default_options`` parameter follows the Silex conventions: you must overwrite it after the provider is registered, not before.

So your PR is not providing consistency. It provides inconsistency by breaking the Silex conventions.

If your issue is related to a config provider registering the config before other providers are registered, this is a broken implementation of the config provider which does not respect the Silex convention, not the opposite.

cgsmith:
Hi @stof, 

The issue is not related to the config provider, that was a misnomer on my part.  I realized that something was not working as I expected when I started to utilize the config service provider.

I think there needs to be discussions on the providers in general as it was not easy for me to quickly grasp that I needed to pass the parameters to the provider.  This logic is non-intuitive and I believe the system should match the user's expectations.  

### Intention of the PR
You mentioned the PR going against Silex conventions, but that is not intended.  I am merely intending the PR to provide another way that is natural to other developers.  It is clear that this PR would be beneficial to a few in the #1176 discussion.  I see no harm as this still allows a consistent experience to other using the conventional method.

What are the negative affects of the PR? Do you have any questions on the positives?

stof:
@cgsmith #1176 was precisely a misunderstanding of the Silex convention. there is strictly nothing to fix in it.

The issue with your PR is that it now uses 2 parameters to represent the runtime value of a parameter and its default value, which does not make sense. The default value is simply set in the parameter and then you can override it to change it.

PR #1215, opened 2015-08-13
Remove duplicate call to add collection

CarsonF:
`doFlush` returns the same `RouteCollection` instance now, so there's no need to add the sub-collection routes to itself.


fabpot:
Thank you @CarsonF.

Issue #1216, opened 2015-08-18
SecurityServiceProvider use CallbackResolver


henrikbjorn:
Currently when using the SecurityServiceProvider and defined the `users` options it can be a closure, but because i dump my config and would like to have all that in a seperate json file. I propose we use the CallbackResolver and checks if it is a Closure when building the service for the firewall.

ref: https://github.com/flint/tacker/issues/8

ragboyjr:
@henrikbjorn do you want to close this or submit something another PR for this?

hkdobrev:
I've cherry-picked @henrikbjorn commits from #1217 and revived his implementation on top of latest master in #1611.

PR #1217, opened 2015-08-20
Allow using a service id when defining user provider

henrikbjorn:
Solves #1216 and through it flint/tacker#8


ragboyjr:
@henrikbjorn, can we close or update this PR, and then #1216?

PR #1223, opened 2015-08-23
Use ArgumentResolver instead of ControllerResolver

henrikbjorn:
This lets us use the standard ControllerResolver and provides
more flexibility by letting users register additional ArgumentResolvers
which could be shared by FrameworkExtra bundle etc.

This obviously needs symfony/symfony#14971 to be merged first.

Additional stuff to add.

- [x] Use the psr-7 bridge for auto converting requests.
- [x] user argument resolver

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,9 @@
             "email": "igor@wiedler.ch"
         }
     ],
+    "repositories" : [
+        { "type" : "vcs", "url" : "http://github.com/WouterJ/symfony" }
+    ],
     "require": {
         "php": ">=5.5.9",
         "pimple/pimple": "~3.0",
@@ -45,7 +48,9 @@
         "twig/twig": ">=1.8.0,<2.0-dev",
         "doctrine/dbal": "~2.2",
         "swiftmailer/swiftmailer": "5.*",
-        "monolog/monolog": "~1.4,>=1.4.1"
+        "monolog/monolog": "~1.4,>=1.4.1",
+        "symfony/symfony" : "dev-pr_11457 as 2.8.0"
+
     },
     "suggest": {
         "symfony/browser-kit": "~2.7",
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -13,8 +13,13 @@ namespace Silex;
 
 use Pimple\Container;
 use Pimple\ServiceProviderInterface;
+use Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolverManager;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\PsrServerRequestArgumentResolver;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestArgumentResolver;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestAttributesArgumentResolver;
 use Symfony\Component\HttpKernel\HttpKernel;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\HttpKernel\TerminableInterface;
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -15,6 +15,9 @@
             "email": "igor@wiedler.ch"
         }
     ],
+    "repositories" : [
+        { "type" : "vcs", "url" : "http://github.com/WouterJ/symfony" }
+    ],
```
henrikbjorn: This needs to be reverted when the PR is merged.

src/Silex/Application.php
```diff
@@ -13,28 +13,35 @@
...
-use Symfony\Component\HttpFoundation\RedirectResponse;
-use Symfony\Component\HttpFoundation\StreamedResponse;
-use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpKernel\TerminableInterface;
 use Symfony\Component\Routing\RouteCollection;
-use Silex\Api\BootableProviderInterface;
-use Silex\Api\EventListenerProviderInterface;
-use Silex\Api\ControllerProviderInterface;
-use Silex\Provider\RoutingServiceProvider;
-use Silex\Provider\KernelServiceProvider;
```
fabpot: Can you revert the reordering as it makes merging 1.x more complex?

src/Silex/Application.php
```diff
@@ -91,12 +98,32 @@ public function __construct(array $values = array())
             return new CallbackResolver($app);
         };
 
+        $this['argument_resolvers'] = function ($app) {
```
fabpot: As we won't bump the requirement of Symfony to 2.8 for Silex 2.0, the code should be compatible with both 2.7 and 2.8.

henrikbjorn:
I think i adressed the different issues :)

src/Silex/Application.php
```diff
@@ -88,7 +94,31 @@ public function __construct(array $values = array())
...
+        };
+
+        $this['argument_resolvers'] = function ($app) {
+            $resolvers = [
+                new RequestArgumentResolver(),
+                new RequestAttributesArgumentResolver(),
+                new ApplicationArgumentResolver($app),
+            ];
+
+            if (interface_exists('Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface')) {
```
stof: you should also check whether Diactoros is available

src/Silex/Application.php
```diff
@@ -88,7 +94,31 @@ public function __construct(array $values = array())
         };
 
         $this['callback_resolver'] = function ($app) {
-            return new CallbackResolver($app);
+            return new CallbackResolver($app, $app['logger']);
+        };
+
+        $this['argument_resolvers'] = function ($app) {
```
stof: this should also be moved once #1224 is merged

src/Silex/ArgumentResolver/ApplicationArgumentResolver.php
```diff
@@ -0,0 +1,38 @@
...
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\ArgumentResolverInterface;
+
+class ApplicationArgumentResolver implements ArgumentResolverInterface
+{
+    private $pimple;
+
+    /**
+     * @param Container $pimple
+     */
+    public function __construct(Container $pimple)
```
stof: given that the resolver is named ApplicationArgumentResolver, should it be typehinted as Application instead ?

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -570,6 +571,14 @@ public function register(Container $app)
 
             $app['validator.validator_service_ids'] = array_merge($app['validator.validator_service_ids'], array('security.validator.user_password' => 'security.validator.user_password_validator'));
         }
+
+        if (class_exists('Symfony\Bundle\SecurityBundle\ArgumentResolver\UserArgumentResolver')) {
```
stof: SecurityBundle is not usable outside the fullstack framework, it depends on FrameworkBundle.
This class should probably be moved to a different location to make it reusable too

stof: and you also need to check for the existence of the interface in HttpKernel first, to be sure that you don't use a recent SecurityBundle version with an older HttpKernel.

src/Silex/ArgumentResolver/ApplicationArgumentResolver.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Silex\ArgumentResolver;
+
+use Pimple\Container;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\ArgumentResolverInterface;
+
+class ApplicationArgumentResolver implements ArgumentResolverInterface
```
stof: this class should have unit tests

PR #1225, opened 2015-08-24
Update travis configuration to use container based infrastructure, al…

henrikbjorn:
…so use cache for composer in order to reduce network usage

Also reformatted yml to use the standard 2 space indentation.

PR #1227, opened 2015-08-24
[1.3] Update travis configuration to use container based infrastructure, al…

henrikbjorn:
…so use cache for composer in order to reduce network usage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: php
 
+sudo: false
+
+cache:
+    directories:
+      - $HOME/.composer/cache
+
 before_script:
     # symfony/*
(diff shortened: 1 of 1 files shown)


henrikbjorn:
The 1.3 edition of #1225

fabpot:
Thank you @henrikbjorn.

.travis.yml
```diff
@@ -1,5 +1,11 @@
 language: php
 
+sudo: false
+
+cache:
+    directories:
+      - $HOME/.composer/cache
```
stof: I suggest caching only the ``/files`` subfolder. this will avoid to invalidate the Travis cache (forcing to persist the updated cache) because of the composer repo cache which will be updated on each build (because Packagist will be updated between all builds)

PR #1231, opened 2015-08-29
changed the app variable in Twig to be an instance of Symfony AppVariable

fabpot:
That's more consistent with Symfony.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -4,6 +4,7 @@ Changelog
 2.0.0 (2015-XX-XX)
 ------------------
 
+* [BC BREAK] The app variable exposed in Twig is now an AppVariable instance
 * [BC BREAK] CSRF has been moved to a standalone provider (``form.secret`` is not available anymore)
 * added support for the Symfony HttpFoundation Twig bridge extension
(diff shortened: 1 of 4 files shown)


fabpot:
And that's probably required for the web profiler.

GromNaN:
Everywhere in Silex,  the app variable is the Silex\Application. Having something different with the same name in templates is inconsistent.

stof:
@fabpot AFAIK, we tried hard to avoid relying on the ``app`` variable in the WebProfilerBundle (it was completely avoided at some point, but I think a feature added another usage again)

jdreesen:
There is a bug in this implementation, because the `AppVariable` is in the `symfony/twig-bridge` which is not necessarily available.

I fixed it with a check for `class_exists('Symfony\Bridge\Twig\AppVariable')` but then I wasn't sure what the behaviour should be if `$app['twig.app_variable']` isn't available and if just falling back to `$twig->addGlobal('app', $app);` would be right, because there would be different behaviour depending on if the twig bridge is loaded or not. Maybe it's okay if we tell people in the documentation, but I don't know if it would be too confusing? This is why I sent no pull request for now.

PR #1232, opened 2015-08-29
updated the Monolog provider to use the newest features of Monolog and the Symfony bridge

fabpot:
@Seldaek If you can have a quick look, that would be much appreciated :)

### DIFF
--- a/src/Silex/Provider/MonologServiceProvider.php
+++ b/src/Silex/Provider/MonologServiceProvider.php
@@ -15,10 +15,12 @@ use Pimple\Container;
 use Pimple\ServiceProviderInterface;
 use Monolog\Formatter\LineFormatter;
 use Monolog\Logger;
-use Monolog\Handler\StreamHandler;
+use Monolog\Handler;
+use Monolog\ErrorHandler;
 use Silex\Application;
 use Silex\Api\BootableProviderInterface;
 use Symfony\Bridge\Monolog\Handler\DebugHandler;
+use Symfony\Bridge\Monolog\Handler\FingersCrossed\NotFoundActivationStrategy;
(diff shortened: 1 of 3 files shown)


src/Silex/Provider/MonologServiceProvider.php
```diff
@@ -77,14 +98,19 @@ public function register(Container $app)
...
         $app['monolog.permission'] = null;
         $app['monolog.exception.logger_filter'] = null;
+        $app['monolog.logfile'] = null;
     }
 
     public function boot(Application $app)
     {
+        if (!$app['debug']) {
+            ErrorHandler::register($app['monolog']);
+        }
```
Seldaek: Do you really mean to only enable this in non-debug mode? I am not familiar enough with silex to know how it behaves internally in debug mode so I'm not sure.

Seldaek:
I think it looks good, except for the fact that monolog.handler (handler) becomes monolog.handlers (array), which is a BC break right? Might be nice to define `monolog.handlers` as `(array) $app['monolog.handler']` in case it is set, but I don't know how much you care about BC.

src/Silex/Provider/MonologServiceProvider.php
henrikbjorn: we have had some errors where the ErrorHandler muted exceptions etc. should be optional imo.

Seldaek: Could that maybe be fixed?

fabpot: Not sure why I did that (I extracted the code from an old project). Let's remove the condition and see if that works for everyone.

PR #1234, opened 2015-08-30
added a provider for VarDumper

fabpot:


henrikbjorn:
already exists https://github.com/flint/Brick/blob/master/src/Provider/DebugServiceProvider.php :+1:

PR #1235, opened 2015-08-30
Fix bug: TwigServiceProvider is broken if FormServiceProvider is registered

andytruong:
form.csrf_provider is refactored to csrf.token_manager.


fabpot:
Thank you @andytruong.

PR #1238, opened 2015-09-01
Remove dependency on Silex\Application from Silex\Api\*

Seldaek:
This allows a CLI Cilex\Application to boot Silex providers.

### DIFF
--- a/src/Silex/Api/BootableProviderInterface.php
+++ b/src/Silex/Api/BootableProviderInterface.php
@@ -11,7 +11,7 @@
 
 namespace Silex\Api;
 
-use Silex\Application;
+use Pimple\Container;
 
 /**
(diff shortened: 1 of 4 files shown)


fabpot:
That does not work. `boot()` is tied with Silex and we should be able to use `Silex\Application` methods. Discussion happened back then and that's the compromise we found.

Seldaek:
OK so the right thing in Cilex would be to skip booting those providers I guess? Or rather extend the Silex\Application? I'm looking for the best way to fix this..

fabpot:
`boot()` is not used that much and IIRC, never needed for Cilex.

Seldaek:
Ok I'll PR that then, thanks

PR #1239, opened 2015-09-01
Remove mandatory dependency on request_stack and some monolog improvements

Seldaek:
This lets everything work with Silex which lacks a request_stack, and I also made a couple changes to the monolog bootstrapping so it's easier to just override the monolog.handler OR monolog.handlers, before you kinda had to redefine both to customize anything.

I also moved the monolog.handler.debug out of the GroupHandler which makes more sense I think because if it is in the group then it falls under the not_found_activation_strategy and therefore even in debug mode you won't get any logs unless you get an error-level record.

### DIFF
--- a/src/Silex/Provider/TranslationServiceProvider.php
+++ b/src/Silex/Provider/TranslationServiceProvider.php
@@ -54,9 +54,11 @@ class TranslationServiceProvider implements ServiceProviderInterface, EventListe
             return $translator;
         };
 
-        $app['translator.listener'] = function ($app) {
-            return new TranslatorListener($app['translator'], $app['request_stack']);
-        };
+        if (isset($app['request_stack'])) {
+            $app['translator.listener'] = function ($app) {
+                return new TranslatorListener($app['translator'], $app['request_stack']);
+            };
+        }
 
         $app['translator.message_selector'] = function () {
             return new MessageSelector();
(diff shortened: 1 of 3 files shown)


Seldaek:
I fixed the build failure. Would be happy to get any feedback here.

stof:
how could Silex 2 lack a RequestStack ? The only case where Silex 1 lacks it is when installing HttpFoundation 2.3 AFAIK, and Silex 2 requires Symfony 2.7+

Seldaek:
Well in Cilex you don't register the HttpKernelServiceProvider and you run into issues.

src/Silex/Provider/TranslationServiceProvider.php
```diff
@@ -73,6 +75,8 @@ public function register(Container $app)
 
     public function subscribe(Container $app, EventDispatcherInterface $dispatcher)
     {
-        $dispatcher->addSubscriber($app['translator.listener']);
+        if (isset($app['request_stack'])) {
```
GromNaN: You should check the existence of `translator.listener`.

GromNaN: You should check the existence of `translator.listener`.

Seldaek: Right, fixed thanks

fabpot:
Is this one a problem for Cilex? 588c52a

Seldaek:
Well no defining debug is alright, that's easy to do if it's not done by default (not sure anymore if it is, but I have it set..), the request_stack on the other hand is kinda problematic because that's something you really don't need in a CLI context

fabpot:
Thank you @Seldaek.

PR #1242, opened 2015-09-02
Remove event type hint as ConsoleEvents are not ResponseEvents

Seldaek:


src/Silex/Provider/SwiftmailerServiceProvider.php
```diff
@@ -99,7 +99,7 @@ public function register(Container $app)
 
     public function subscribe(Container $app, EventDispatcherInterface $dispatcher)
     {
-        $onTerminate = function (PostResponseEvent $event) use ($app) {
```
fabpot: Can you add a comment so that we do not inadvertently re-add the type hint?

Seldaek: Done

fabpot:
Thank you @Seldaek.

PR #1243, opened 2015-09-02
Verify that the spool is enabled and an actual spool before flushing it

Seldaek:
At least the use_spool check is necessary, the other third check IMO is a nice safeguard because for example I had redefined the swiftmailer.spooltransport to be swiftmailer.transport as a workaround (I don't think use_spool existed back then).


fabpot:
should probably be merged in 1.3.

PR #1245, opened 2015-09-02
Fix TwigServiceProvider if no Symfony\Bridge\Twig available

jdreesen:
As already described in https://github.com/silexphp/Silex/pull/1231#issuecomment-136828673 the change introduced an error if the `symfony/twig-bridge` is not loaded: `Fatal error: Class 'Symfony\Bridge\Twig\AppVariable' not found in \vendor\silex\silex\src\Silex\Provider\TwigServiceProvider.php on line 43`

My first try for a solution was to only add the `AppVariable` if the twig-bridge is available and just fall back to adding `$app` otherwise. But I'm not sure if this is the desired behaviour, so I'm curious what you think about this.

The documentation has to be adapted once it's clear how the behaviour should be.

### DIFF
--- a/src/Silex/Provider/TwigServiceProvider.php
+++ b/src/Silex/Provider/TwigServiceProvider.php
@@ -39,18 +39,20 @@ class TwigServiceProvider implements ServiceProviderInterface
         $app['twig.path'] = array();
         $app['twig.templates'] = array();
 
-        $app['twig.app_variable'] = function ($app) {
-            $var = new AppVariable();
-            if (isset($app['security.token_storage'])) {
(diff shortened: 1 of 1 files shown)


fabpot:
Not sure what to do here as there is another comment somewhere saying that this is a significant BC break as well with 1.x as `$app` was the full Silex application. That's not really a problem as we break BC in 2.0, but as `app` refers to the Silex App everywhere, that might be confusing as well.

henrikbjorn:
Could we rename it? Otherwise we just have to be very precise about what `app` is refering in the documentation.

henrikbjorn:
Could the global `app` be replaced with a group of extensions?

PR #1247, opened 2015-09-07
decoupled the exception handler from HttpKernelServiceProvider

fabpot:


henrikbjorn:
I am curious, what is the reasoning for doing this?

fabpot:
Pure semantic. Same reasoning as for the other providers: they only do one thing and the exception handler has nothing to do with the other services of the provider.

GromNaN:
:+1:

stof:
:+1:

PR #1253, opened 2015-09-15
reverted app Twig variable change, introduced a new global one instead

fabpot:
fixes #1245


fabpot:
Note that using `app` in a Twig template is deprecated.

PR #1256, opened 2015-09-15
Add option to specify another route param for converters

mathroc:
As said in #1211 this simplify renaming a route parameter to another controller variable name, this is reduce boilerplate a lot, especially when using existing providers as converters:

``` php
$app->get('/user/:username', function(UserInterface $user) {
    // ...
})->convert('user', 'provider.user:loadUserByUsername', 'username');
```

instead of:

``` php
$app->get('/user/:username', function(UserInterface $user) {
    // ...
})->convert('user', function($_, Request $request) use ($app) {
    return $app["provider.user"]->loadUserByUsername($request->get('username'));
});
```

### DIFF
--- a/src/Silex/EventListener/ConverterListener.php
+++ b/src/Silex/EventListener/ConverterListener.php
@@ -49,10 +49,11 @@ class ConverterListener implements EventSubscriberInterface
         $request = $event->getRequest();
         $route = $this->routes->get($request->attributes->get('_route'));
         if ($route && $converters = $route->getOption('_converters')) {
-            foreach ($converters as $name => $callback) {
+            foreach ($converters as $name => list($callback, $originalName)) {
                 $callback = $this->callbackResolver->resolveCallback($callback);
+                $originalValue = $request->attributes->get($originalName);
 
-                $request->attributes->set($name, call_user_func($callback, $request->attributes->get($name), $request));
+                $request->attributes->set($name, call_user_func($callback, $originalValue, $request));
             }
         }
     }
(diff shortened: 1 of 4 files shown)


wI2L:
+1

henrikbjorn:
Isn't this possible but just having the `$user` be sent to the converter?

``` php
$app->get('/user/:user', function(UserInterface $user) {
    // ...
})->convert('user', 'provider.user:loadUserByUsername');
```

mathroc:
@henrikbjorn yes you can, this PR is about make it easy to have a different name in the route and the controller, eg: username -> User, userid -> User or slug -> Post

henrikbjorn:
:-1: I really don't see the benefit of this. It is already possible to do what you want (bascially).

mathroc:
what is it you don't see the benefit of ? having a different name in the route and controller ? or to have another way to it with less boilerplate code ?

henrikbjorn:
it's Magic and i dont see the benefit of using a different parameter name and a different variable name in the route callback.

mathroc:
Hi,

about using a different parameter name in the route and the controller, I still think it's valuable, it makes route more explicit about their parameters (is `:user` the user id, its login or its email?). and it comes with very little maintenance cost.
anyone feeling the same ?

srsbiz:
+1

mathroc:
@fabpot any chance to get this into 2.0 ? or should it be closed ?

fabpot:
I think it's safe to close as I'm not convinced this is worth it (Symfony has the same issue as well).

PR #1266, opened 2015-10-03
Fix error never removed from session in SecurotyProvider

Savageman:
When $message is different than the $error variable, the wrong session key is deleted.


fabpot:
Thank you @Savageman.

dereuromark:
https://github.com/silexphp/Silex/commit/2509b5ac001a16beb7a380204a3d577701280222

fabpot:
1.3.5 released, sorry for the delay.

dereuromark:
Thx! That was quick :)

PR #1271, opened 2015-10-20
Don't mute monolog.permission in MonologServiceProvider

fliak:
MonologServiceProvider mute permission settings placed by config file.


stof:
-1 for that. This does not match Silex conventions. Parameter customizations must be done **after** registering providers

fliak:
Understand, thank you!

PR #1283, opened 2015-11-13
fix typo

nicoder:

PR #1284, opened 2015-11-13
fix typo

nicoder:


SpacePossum:
:+1:

GromNaN:
You could have made only 1 PR for all #1283 and #1282.

nicoder:
yes, with hindsight I would have, I guess I did not expect to find more.
Next time I should wait until I am done reading before sending the PRs, and juste send one if they are related.

GromNaN:
You can (and should !) open a PR as soon as you find the first typo, but if you find another one just after you can add commits to the same branch, the PR will be automatically updated.

Regardless, thank you for fixing the doc.

nicoder:
Oh, thanks, I hadn't thought about that, good idea.
I just clicked on 'Edit this' on the website and edited through the Github web UI,
which created a new branch every time.
I could've juste done that on the same branch on my fork in the web UI.

fabpot:
Good catch, thanks @nicoder.

PR #1285, opened 2015-11-13
s/middlewares/middleware

nicoder:
Feel free to discard this PR,
but reading 'Middlewares' with an 's' feels jarring to me,
I never saw it written this way before.

For uses without the 's' see :
http://english.stackexchange.com/questions/257120/middleware-vs-middlewares
https://en.wikipedia.org/wiki/Middleware
http://expressjs.com/guide/using-middleware.html

I did find uses of 'middlewares' in these pages, so, again, feel free to discard :
http://guides.rubyonrails.org/rails_on_rack.html
https://blog.engineyard.com/2015/understanding-rack-apps-and-middleware

But for example it appears just twice with an 's' in this stack overflow page, and many more times without the 's' :
http://stackoverflow.com/questions/2256569/what-is-rack-middleware

### DIFF
--- a/doc/middlewares.rst
+++ b/doc/middlewares.rst
@@ -1,18 +1,18 @@
-Middlewares
+Middleware
 ===========
 
 Silex allows you to run code, that changes the default Silex behavior, at
-different stages during the handling of a request through *middlewares*:
+different stages during the handling of a request through *middleware*:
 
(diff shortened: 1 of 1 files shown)


ragboyjr:
are going to do anything with this?

fabpot:
Thank you @nicoder.

PR #1289, opened 2015-11-26
Remove slash from FallbackResource

JeroenDeDauw:
Apache config is not my thing, so this might not be correct. However, what I can tell is that it does not work with the slash for me, and that it does without the slash.


hkdobrev:
:+1:

fabpot:
Thank you @JeroenDeDauw.

Issue #1262, opened 2015-09-26
How to group routes


lautiamkok:
Can I group my routes in Silex just as Slim framework?

For instance, in Slim,
```

// API group
$app->group('/api', function () use ($app) {

    // Admin group
    $app->group('/admin', function () use ($app) {

        // Get book with ID
        $app->get('/books/:id', function ($id) use ($app) {
            // Update book identified by $id.
        });

        // Update book with ID
        $app->put('/books/:id', function ($id) use ($app) {
            // Update book identified by $id.
        });

        // Delete book with ID
        $app->delete('/books/:id', function ($id) use ($app) {
            // Update book identified by $id.
        });

    });

});
```

Is it possible with Silex?

hkdobrev:
@lauthiamkok You'd want to read the documentation on organizing controllers and mounting: http://silex.sensiolabs.org/doc/organizing_controllers.html

PR #1290, opened 2015-11-28
Route groups

talyssonoc:
Solves the proposed implementation at https://github.com/silexphp/Silex/issues/1262

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -295,6 +295,19 @@ class Application extends \Pimple implements HttpKernelInterface, TerminableInte
     }
 
     /**
+     * Creates a mounting point and creates a callable to mount controllers under it.
+     *
+     * @param string $prefix The mounting point prefix
+     * @param mixed  $to     Callback that creates the controllers under the mounting point
+     *
+     * @return Controller
+     */
+    public function group($prefix, $to = null)
+    {
+        return $this['controllers']->group($prefix, $to);
+    }
+
+    /**
      * Adds an event listener that listens on the specified events.
      *
      * @param string   $eventName The event to listen on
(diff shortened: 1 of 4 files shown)


talyssonoc:
@fabpot Finally! After a lot of errors with merge (I've forgot I created the branch from `1.3`, so when I tried to merge with `master` I got a lot of conflicts and extra files... twice)

fabpot:
@talyssonoc Thanks for your work on this. Several quick comments:

 * we should also have some docs about this new feature
 * as this is a new feature, it should be based on the master branch, not 1.3 (sorry about that)
 * I'm not sure about the `group` name (it was not obvious for me what it refers to)

ragboyjr:
@talyssonoc, @fabpot, what if instead of using the word `group`, we just overload the `mount` function. This works exactly like mount does, we'd just let it take in a function as well.

```php
$app->mount('/api', function ($api) use ($app) {
    // Admin group
    $api->mount('/admin', function ($admin) use ($app) {
        // Get book with ID
        $admin->get('/books/:id', function ($id) use ($app) {
            // Update book identified by $id.
        });

        // Update book with ID
        $admin->put('/books/:id', function ($id) use ($app) {
            // Update book identified by $id.
        });

        // Delete book with ID
        $admin->delete('/books/:id', function ($id) use ($app) {
            // Update book identified by $id.
        });
    });
});
```

fabpot:
@ragboyjr Looks like a fantastic idea!

ragboyjr:
@talyssonoc, do you plan on making those changes, or should I make a separate PR?

fabpot:
Closing as #1345 looks better to me.

PR #1292, opened 2015-12-05
first changes to try get legacy provider up and running within silex 2.0

dominikzogg:
Hello @stof 

I tried to create a legacy provider support as we discus about today. Its in a early stage and not working correctly at the moment.

One problem for example would be provider which tried to create new \Pimple instances.

https://github.com/silexphp/Silex/blob/1.3/src/Silex/Provider/DoctrineServiceProvider.php#L65

Regards @dominikzogg

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -99,7 +99,7 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
      * Registers a legacy service provider.
      *
      * @param LegacyServiceProviderInterface $provider A LegacyServiceProviderInterface instance
-     * @param array $values An array of values that customizes the provider
+     * @param array                          $values An array of values that customizes the provider
      *
      * @return Application
      */
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -49,7 +49,7 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
     const LATE_EVENT = -512;
 
     protected $providers = array();
-    protected $legacyProvider = false;
+    private $legacyProvider = false;
     protected $booted = false;
 
     /**
@@ -102,19 +102,28 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
      * @param array                          $values   An array of values that customizes the provider
      *
      * @return Application
+     *
+     * @throws \Exception
      */
     public function registerLegacy(LegacyServiceProviderInterface $provider, array $values = array())
     {
         $this->providers[] = $provider;
-        $this->legacyProvider = true;
 
-        $provider->register($this);
+        try {
+            $this->legacyProvider = true;
 
-        foreach ($values as $key => $value) {
-            $this[$key] = $value;
+            $provider->register($this);
+
+            foreach ($values as $key => $value) {
+                $this[$key] = $value;
+            }
+
+            $this->legacyProvider = false;
+        } catch (\Exception $e) {
+            $this->legacyProvider = false;
+
+            throw $e;
         }
-
-        $this->legacyProvider = false;
 
         return $this;
     }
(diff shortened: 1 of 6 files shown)


dominikzogg:
@stof ok its working now, which means the old monolog provider is usable within silex 2. but i had to change some code withinin its tests, cause log level changed, and override some services needs to be done in a pimple 3 way.

src/Silex/Application.php
```diff
@@ -47,6 +49,7 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
     const LATE_EVENT = -512;
 
     protected $providers = array();
+    protected $legacyProvider = false;
```
stof: this should be private, to avoid letting child classes messing with it

src/Silex/Application.php
```diff
@@ -93,6 +96,30 @@ public function register(ServiceProviderInterface $provider, array $values = arr
...
+        $this->providers[] = $provider;
+        $this->legacyProvider = true;
+
+        $provider->register($this);
+
+        foreach ($values as $key => $value) {
+            $this[$key] = $value;
+        }
+
+        $this->legacyProvider = false;
```
stof: you need to reset this also when an exception happens during the registration

src/Silex/Application.php
```diff
@@ -107,6 +134,16 @@ public function boot()
         $this->booted = true;
 
         foreach ($this->providers as $provider) {
+            if ($provider instanceof LegacyServiceProviderInterface) {
+                $this->legacyProvider = true;
+
+                $provider->boot($this);
+
+                $this->legacyProvider = false;
```
stof: same issue here

src/Silex/Application.php
```diff
@@ -503,4 +540,65 @@ public function terminate(Request $request, Response $response)
...
+            parent::offsetSet($id, parent::factory($value));
+
+            return;
+        }
+
+        parent::offsetSet($id, $value);
+    }
+
+    /**
+     * Returns a closure that stores the result of the given service definition
```
stof: wrong description

src/Silex/ServiceProviderInterface.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+namespace Silex;
+
+/**
+ * Interface that all Silex service providers must implement.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ * 
+ * @deprecated
```
stof: missing deprecation message

src/Silex/Legacy/SharedService.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Silex\Legacy;
+
+class SharedService
```
stof: the class should be marked as internal

stof:
Is BC provided for ``extend`` too ?

src/Silex/Legacy/SharedService.php
stof: and you may want to make it invokable to be more BC

GromNaN:
It might be a stupid question, but why not upgrade existing service providers for Silex2 / Pimple3 ?

All service providers of the core of Silex have been upgraded quite easily. What are you trying to use that requires so much effort that you find it easier to modify both Pimple and Silex than that code ?

dominikzogg:
@GromNaN the goal was to provide an easy to use solutions for old providers. At the moment its only a proof of concept.

dominikzogg:
@stof yeah, bc of extend should be given

dominikzogg:
@stof why do think that SharedService needs to be invokable? its only used as helper within Application class

dominikzogg:
@fabpot the problem, that i see with such a solution is, that every legacy provider which creates its own pimple instances as for example dbal provider does, would not work. Which means there would be pimple changes too, and it get very ugly.

https://github.com/silexphp/Silex/blob/1.3/src/Silex/Provider/DoctrineServiceProvider.php#L65

fabpot:
@dominikzogg Indeed :( Perhaps you were right when we discussed that topic; it's maybe not worth it.

dominikzogg:
@fabpot i close this PR and the one against pimple

Issue #1260, opened 2015-09-24
Support the new Symfony Guard component


fabpot:

henrikbjorn:
Is there any reference docs for Symfony on this?

andytruong:
I think it's https://github.com/symfony/symfony/pull/14673

ChrisRiddell:
Will this be part of Silex 2?

fabpot:
If someone wants to work on it. I won't have time in the next few weeks.

mathroc:
Could it come in Silex 2.1 instead of waiting for it for 2.0?

PR #1296, opened 2015-12-17
Add support for the Guard component to the SecurityServiceProvider

GromNaN:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1260
| License       | MIT
| Doc PR        | included

Services configuration are extracted from the SecurityBundle [`guard.xml`](https://github.com/symfony/symfony/blob/3.0/src/Symfony/Bundle/SecurityBundle/Resources/config/guard.xml) and [`GuardAuthenticationFactory`](https://github.com/symfony/symfony/blob/3.0/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/GuardAuthenticationFactory.php)

Usage is quite simple, the `guard` type can be configured in the firewall like other ones.

```php
$app['app.authenticator'] = function ($app) {
    return new Authenticator();
};

$app->register(new Silex\Provider\SecurityServiceProvider(), [
    'security.firewalls' => [
        'main' => [
            'pattern' => '^/admin',
            'guard' => [
                'authenticators' => [
                    'app.authenticator'
                ]
            ]
        ]
    ]
]);
```

### DIFF
--- a/doc/cookbook/guard_authentication.rst
+++ b/doc/cookbook/guard_authentication.rst
@@ -1,0 +1,40 @@
+How to Create a Custom Authentication System with Guard
+=======================================================
+
+TODO: This cookbook will be inspired by the Symfony cookbook.
+http://symfony.com/doc/current/cookbook/security/guard-authentication
+
+Step 1) Create the Authenticator Class
+--------------------------------------
+
+TODO: Same as the Symfony cookbook, but Doctrine ORM is not recommanded with Silex.
+
+Step 2) Configure the Authenticator
+-----------------------------------
+
+To finish this, register the class as a service:
+
+.. code-block:: php
+
+    $app['app.token_authenticator'] = function ($app) {
+        return new App\Security\TokenAuthenticator();
+    }
+
+
+Finally, configure your firewalls key to use this authenticator:
+
+.. code-block:: php
+
+    $app['security.firewalls'] => array(
+        'main' => array(
+            'anonymous' => true,
+
+            'guard' => array(
+                'authenticators' => array(
+                    'app.token_authenticator'
+                ),
+            ),
+        ),
+    );
+
+
--- a/doc/cookbook/guard_authentication.rst
+++ b/doc/cookbook/guard_authentication.rst
@@ -33,8 +33,13 @@ Finally, configure your firewalls key to use this authenticator:
                 'authenticators' => array(
                     'app.token_authenticator'
                 ),
+
+                // Using more than 1 authenticator, you must specify
+                // which one is used as entry point.
+                'entry_point' => 'app.token_authenticator',
             ),
         ),
     );
 
+Note: You can use many authenticators, they are executed by if the order they are configured.
 
--- a/doc/cookbook/guard_authentication.rst
+++ b/doc/cookbook/guard_authentication.rst
@@ -1,13 +1,98 @@
 How to Create a Custom Authentication System with Guard
 =======================================================
 
-TODO: This cookbook will be inspired by the Symfony cookbook.
-http://symfony.com/doc/current/cookbook/security/guard-authentication
+Whether you need to build a traditional login form, an API token authentication system or you need to integrate with some proprietary single-sign-on system, the Guard component can make it easy... and fun!
+
+In this example, you'll build an API token authentication system and learn how to work with Guard.
 
 Step 1) Create the Authenticator Class
 --------------------------------------
 
-TODO: Same as the Symfony cookbook, but Doctrine ORM is not recommanded with Silex.
+Suppose you have an API where your clients will send an X-AUTH-TOKEN header on each request. This token is composed of the username followed by a password, separated by a colon. Your job is to read this, find the associated user (if any) and check the password.
+
+To create a custom authentication system, just create a class and make it implement GuardAuthenticatorInterface. Or, extend the simpler AbstractGuardAuthenticator. This requires you to implement six methods:
+
+.. code-block:: php
+
+    <?php
+
+    namespace App\Security;
+
+    use Symfony\Component\HttpFoundation\Request;
+    use Symfony\Component\HttpFoundation\JsonResponse;
+    use Symfony\Component\Security\Core\User\UserInterface;
+    use Symfony\Component\Security\Core\User\UserProviderInterface;
+    use Symfony\Component\Security\Guard\AbstractGuardAuthenticator;
+    use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+    use Symfony\Component\Security\Core\Exception\AuthenticationException;
+
+    class TokenAuthenticator extends AbstractGuardAuthenticator
+    {
+        public function getCredentials(Request $request)
+        {
+            if (!$token = $request->headers->get('X-AUTH-TOKEN')) {
+                return;
+            }
+
+            list($username, $secret) = explode(':', $token, 2);
+
+            return array(
+                'username' => $username,
+                'secret' => $secret,
+            );
+        }
+
+        public function getUser($credentials, UserProviderInterface $userProvider)
+        {
+            return $userProvider->loadUserByUsername($credentials['username']);
+        }
+
+        public function checkCredentials($credentials, UserInterface $user)
+        {
+            // check credentials - e.g. make sure the password is valid
+            // no credential check is needed in this case
+
+            // return true to cause authentication success
+            return $user->getPassword() === $credentials['secret'];
+        }
+
+        public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
+        {
+            // on success, let the request continue
+            return;
+        }
+
+        public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
+        {
+            $data = array(
+                'message' => strtr($exception->getMessageKey(), $exception->getMessageData()),
+
+                // or to translate this message
+                // $this->translator->trans($exception->getMessageKey(), $exception->getMessageData())
+            );
+
+            return new JsonResponse($data, 403);
+        }
+
+        /**
+         * Called when authentication is needed, but it's not sent
+         */
+        public function start(Request $request, AuthenticationException $authException = null)
+        {
+            $data = array(
+                // you might translate this message
+                'message' => 'Authentication Required',
+            );
+
+            return new JsonResponse($data, 401);
+        }
+
+        public function supportsRememberMe()
+        {
+            return false;
+        }
+    }
+
 
 Step 2) Configure the Authenticator
 -----------------------------------
@@ -21,14 +106,12 @@ To finish this, register the class as a service:
     }
 
 
-Finally, configure your firewalls key to use this authenticator:
+Finally, configure your `security.firewalls` key to use this authenticator:
 
 .. code-block:: php
 
     $app['security.firewalls'] => array(
         'main' => array(
-            'anonymous' => true,
-
             'guard' => array(
                 'authenticators' => array(
                     'app.token_authenticator'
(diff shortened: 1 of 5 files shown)


GromNaN:
This PR is ready to be reviewed.

doc/cookbook/guard_authentication.rst
```diff
@@ -0,0 +1,150 @@
...
+                // 'entry_point' => 'app.token_authenticator',
+            ),
+            'users' => array(
+                'victoria' => array('ROLE_USER', 'randomsecret'),
+            ),
+        ),
+    );
+
+.. note::
+    You can use many authenticators, they are executed by if the order they are configured.
```
davedevelopment: > by if

s/if//

davedevelopment:
A tentative :+1: from me, I've only read the code, not run anything :)

GromNaN:
Maybe @weaverryan could have a look.

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -148,12 +151,14 @@ public function register(Container $app)
...
-        foreach (array('logout', 'pre_auth', 'form', 'http', 'remember_me', 'anonymous') as $type) {
+        foreach (array('logout', 'pre_auth', 'guard', 'form', 'http', 'remember_me', 'anonymous') as $type) {
             $entryPoint = null;
             if ('http' === $type) {
                 $entryPoint = 'http';
             } elseif ('form' === $type) {
                 $entryPoint = 'form';
+            } elseif ('guard' === $type) {
+                $entryPoint = 'guard';
             }
```
weaverryan: :+1: this is consistent with the full-stack: we don't replace the entry point of other auth methods, just to be safe

doc/cookbook/guard_authentication.rst
```diff
@@ -0,0 +1,150 @@
...
+=======================================================
+
+Whether you need to build a traditional login form, an API token authentication system or you need to integrate with some proprietary single-sign-on system, the Guard component can make it easy... and fun!
+
+In this example, you'll build an API token authentication system and learn how to work with Guard.
+
+Step 1) Create the Authenticator Class
+--------------------------------------
+
+Suppose you have an API where your clients will send an X-AUTH-TOKEN header on each request. This token is composed of the username followed by a password, separated by a colon. Your job is to read this, find the associated user (if any) and check the password.
```
weaverryan: I think the standard is to add line breaks after about the 72nd character.

Also, you should give an example if the header:

```rst
the username followed by a password, separated by a colon (e.g. ``X-AUTH-TOKEN: coolguy:awesomepassword``)
```

doc/cookbook/guard_authentication.rst
```diff
@@ -0,0 +1,150 @@
...
+            return $userProvider->loadUserByUsername($credentials['username']);
+        }
+
+        public function checkCredentials($credentials, UserInterface $user)
+        {
+            // check credentials - e.g. make sure the password is valid
+            // no credential check is needed in this case
+
+            // return true to cause authentication success
+            return $user->getPassword() === $credentials['secret'];
```
weaverryan: Is the `secret` (the password) supposed to be encoded when the user sends it? If it's plain-text, then we would need to encode that secret to compare it with the encoded password, right?

doc/cookbook/guard_authentication.rst
weaverryan: yep :)

weaverryan:
Thanks @GromNaN for this (and the ping!). I left some comments, but it looks great. I also didn't try it locally... but your functional test speaks for itself.

Let's get those tweaks made and get this guy merged finally :)

tests/Silex/Tests/Provider/SecurityServiceProviderTest/TokenAuthenticator.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public function getUser($credentials, UserProviderInterface $userProvider)
+    {
+        return $userProvider->loadUserByUsername($credentials['username']);
+    }
+
+    public function checkCredentials($credentials, UserInterface $user)
+    {
+        return $user->getPassword() === $credentials['secret'];
```
GromNaN: I'll add a comment tailling this is not a safe way of checking password.

vicmosin:
Have anyone an idea when this feature comes to release?

weaverryan:
I'll review this PR again soon, including testing it locally.

doc/cookbook/guard_authentication.rst
```diff
@@ -0,0 +1,150 @@
...
+Step 2) Configure the Authenticator
+-----------------------------------
+
+To finish this, register the class as a service:
+
+.. code-block:: php
+
+    $app['app.token_authenticator'] = function ($app) {
+        return new App\Security\TokenAuthenticator();
+    }
```
weaverryan: missing `;` at the end - `};`

doc/cookbook/guard_authentication.rst
```diff
@@ -0,0 +1,150 @@
...
+
+        public function getUser($credentials, UserProviderInterface $userProvider)
+        {
+            return $userProvider->loadUserByUsername($credentials['username']);
+        }
+
+        public function checkCredentials($credentials, UserInterface $user)
+        {
+            // check credentials - e.g. make sure the password is valid
+            // no credential check is needed in this case
```
weaverryan: probably this second bullet should be removed :)

doc/cookbook/guard_authentication.rst
weaverryan: This should be:

```php
        $encoder = $this->encoderFactory->getEncoder($user);

        return $encoder->isPasswordValid(
            $user->getPassword(),
            $credentials['secret'],
            $user->getSalt()
        );
```

With a constructor:

```php
    private $encoderFactory;

    public function __construct(EncoderFactoryInterface $encoderFactory)
    {
        $this->encoderFactory = $encoderFactory;
    }
```

And the service registration needs to be updated:

```php
$app['app.token_authenticator'] = function ($app) {
    return new App\Security\TokenAuthenticator($app['security.encoder_factory']);
};
```

doc/cookbook/guard_authentication.rst
```diff
@@ -0,0 +1,150 @@
...
+            'guard' => array(
+                'authenticators' => array(
+                    'app.token_authenticator'
+                ),
+
+                // Using more than 1 authenticator, you must specify
+                // which one is used as entry point.
+                // 'entry_point' => 'app.token_authenticator',
+            ),
+            'users' => array(
```
weaverryan: I'd like to add a comment above this:

```php
// configure where your users come from. Hardcode them, or load them from somewhere
// http://silex.sensiolabs.org/doc/providers/security.html#defining-a-custom-user-provider
```

doc/cookbook/guard_authentication.rst
```diff
@@ -0,0 +1,150 @@
...
+                'victoria' => array('ROLE_USER', 'randomsecret'),
+            ),
```
weaverryan: I'd also like:

```php
// 'anonymous' => true
```

weaverryan:
@GromNaN I just tested this locally and added some corrections so we can get this guy merged! Also:

A) We should link to the cookbook entry form the security provider chapter
B) Is this meant to be only for Silex 2? The PR is against the master branch - if we want to be able to merge into Silex 1, it should be rebased against 1.3. That can be done on merge, I just don't know if there are actual code changes that would need to happen (there probably are) to be compatible with 1.3.

Also, in case it helps, I have a working project using your code here: https://github.com/weaverryan/silex-guard-test-example/tree/finished. You just need to delete `vendor/silex/silex` and replace it with your fork on the `guard` branch.

Thanks!

GromNaN:
Thanks, I'll check all that later today.

doc/cookbook/guard_authentication.rst
GromNaN: What about `password_verify` ? That's a lot simpler.

GromNaN:
@weaverryan I patched the doc.

Silex 1 and 2 have strong incompatibilities (pimple version, service provider definition) ...
I can backport this PR to Silex 1.3, but the code and the doc will be different between Silx 1.3 and 2.

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -420,6 +430,29 @@ public function register(Container $app)
...
+                if (!isset($app['security.authentication.guard_handler'])) {
+                    $app['security.authentication.guard_handler'] = new GuardAuthenticatorHandler($app['security.token_storage'], $app['dispatcher']);
+                }
+
+                $authenticators = array();
+                foreach ($options['authenticators'] as $authenticatorId) {
+                    $authenticators[] = $app[$authenticatorId];
+                }
+
+                $class = isset($options['listener_class']) ? $options['listener_class'] : 'Symfony\\Component\\Security\\Guard\\Firewall\\GuardAuthenticationListener';
```
stof: does the listener class need to be configurable here ?

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -545,7 +578,24 @@ public function register(Container $app)
...
-        $app['security.authentication_provider.dao._proto'] = $app->protect(function ($name) use ($app) {
+        $app['security.entry_point.guard._proto'] = $app->protect(function ($name, array $options) use ($app) {
+            if (isset($options['entry_point'])) {
+                // if it's configured explicitly, use it!
+                return $app[$options['entry_point']];
+            }
+            $authenticatorIds = $options['authenticators'];
+            if (count($authenticatorIds) == 1) {
+                // if there is only one authenticator, use that as the entry point
+                return $app[array_shift($authenticatorIds)];
```
stof: using array_shift to get the first argument creates a reference mismatch here on PHP 5.x (PHP 7 does not have reference mismatches anymore) because the function takes a reference as argument. This forces to duplicate the array.
PHP 7 will also need to duplicate the array as you alter it.

Getting the first element without changing the array would be better.

GromNaN: I took this code from [`GuardAuthenticationFactory`](https://github.com/symfony/symfony/blob/v3.0.4/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/GuardAuthenticationFactory.php#L112-L115)

What about using `current($authenticatorIds)` instead ?

src/Silex/Provider/SecurityServiceProvider.php
GromNaN: Not really. I replicated what has been done on others.

src/Silex/Provider/SecurityServiceProvider.php
SpacePossum: `current` might not work if the pointer is pointing to the end of the array. Since we know the count of the array is one wouldn't `reset` be safe and fast (i.e. no copy and no risk)?

weaverryan: In practice, the way it's written should be fine, right? Specifically, `$authenticatorIds = $options['authenticators'];` makes a copy of the array, and then this new copy is only used for this one purpose.

But, +1 also for `reset()` as a solution.

mathroc:
FWIW, I've been using in it in a some projects for a few weeks to implements token based and session authentication with success :)

thx @GromNaN !

fabpot:
Thank you @GromNaN.

src/Silex/Provider/SecurityServiceProvider.php
stof: @weaverryan ``$authenticatorIds = $options['authenticators']`` does not copy the array, as PHP performs copy-on-write.

``reset`` would cause the same issue, as it also takes a reference, and so causes a reference mismatch.

SpacePossum: since the key of the first (and only in this case) element is unknown, how to get the first element without the mismatch?

PR #1300, opened 2016-01-04
Replace binary gender examples with something that is actually single-choice

igorwwwwwwwwwwwwwwwwwwww:
In order to include people who do not fit into the gender binary, gender fields really should be a free-form text field. As such, this patch changes the `choices` example to use a billing plan instead.

# References

* http://gender.wikia.com/wiki/Non-binary
* https://en.wikipedia.org/wiki/Genderqueer
* https://modelviewculture.com/pieces/the-argument-for-free-form-input
* https://www.facebook.com/

### DIFF
--- a/doc/providers/form.rst
+++ b/doc/providers/form.rst
@@ -91,8 +91,8 @@ example::
         $form = $app['form.factory']->createBuilder('form', $data)
             ->add('name')
             ->add('email')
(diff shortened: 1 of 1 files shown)


davedevelopment:
:+1:

fabpot:
Thank you @igorwwwwwwwwwwwwwwwwwwww.

PR #1305, opened 2016-01-15
Request hash with monolog

hasumedic:
I recently found myself in the need of adding a request hash into the logs of a Silex project. I could not find anywhere a cookbook or documentation on how to do this, so I documented the way I solved it.

It's actually pretty similar (if not the same) solution as proposed in the Symfony docs:
http://symfony.com/doc/current/cookbook/logging/monolog.html#adding-a-session-request-token
I've simplified it the best I could.

This should help the next poor soul trying to achieve this.

I'm asking this to be merged into 1.3 since it's only documentation and not a new feature.

Thanks

### DIFF
--- a/doc/cookbook/monolog_request_hash.rst
+++ b/doc/cookbook/monolog_request_hash.rst
@@ -1,0 +1,84 @@
+Adding a request hash using Monolog
+===================================
+
+Often times is useful to add some sort of hash or identifier to each request
+in the logs so it's easier to aggregate or filter its lines with tools like grep
+or more advanced ones like `Logstash
+<https://www.elastic.co/products/logstash>`_.
+
+This is something simple to achieve with Monolog by means of its Formatters and
+Processors.
+
+First up, we need to create a new Monolog Processor which is going to slightly
+transform the record to be logged before it actually gets logged. At this stage,
+we'll create the request hash if it hasn't been created already.
+
+.. code-block:: php
+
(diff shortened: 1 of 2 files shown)


doc/cookbook/monolog_request_hash.rst
```diff
@@ -0,0 +1,84 @@
+Adding a request hash using Monolog
+==============================
```
fabpot: RST is very strict, so you need to add the missing `=` at the end

hasumedic: Sorry about that. Please let me know if there's something else that needs tweaking before it can be merged ;)

doc/cookbook/monolog_request_hash.rst
```diff
@@ -0,0 +1,84 @@
...
+This is something simple to achieve with Monolog by means of its Formatters and
+Processors.
+
+First up, we need to create a new Monolog Processor which is going to slightly
+transform the record to be logged before it actually gets logged. At this stage,
+we'll create the request hash if it hasn't been created already.
+
+.. code-block:: php
+
+    final class TokenProcessor
```
tgallice: You can use the `UidProcessor` instead of create a new one.

fabpot: I agree

fabpot:
Closing as there is a better solution as explained by @tgallice

Issue #1302, opened 2016-01-07
Silex FormServiceProvider could not load type “form” when using Symfony 3 components


bestattendance:
When trying to build a form I receive this error:

    InvalidArgumentException in FormRegistry.php line 87: Could not load type "form"

I've registered the FormServiceProvider, TranslationServicerProvider, and ValidatorServiceProvider.

Here is the relevant part of my code:

        $this->_form = $this->_app['form.factory']->createBuilder('form', $this->_map())
        ->add('firstName', 'text', [
            'constraints' => [new Assert\NotBlank()]
        ])
        ->add('lastName', 'text', [
            'constraints' => [new Assert\NotBlank()]
        ])
        ->add('email', 'text', [
            'constraints' => [new Assert\Email()]
        ])
        ->getForm();

Here are version numbers of the related components that I'm using:

silex/silex v1.3.5
symfony/security-core    v3.0.1
symfony/security-csrf v3.0.1
symfony/translation v3.0.1
symfong/twig-bridge v3.0.1
symfony/validator v3.0.1
symfony/form v3.0.1

Downgrading all symfony components to 2.8.1 solved the problem.

bestattendance:
Ref: http://stackoverflow.com/questions/34663227/silex-formserviceprovider-could-not-load-type-form

hellboy1975:
I can confirm this problem.  Downgrading the Symfony components was the only option I could find that worked.

HeahDude:
This is not a `FormServiceProvider` issue.
It's about one of the few bc break introduced in version 3 of symfony components. You can get this working by changing the deprecated alias `text` by `TextType::class` :

```php
use Symfony\Component\Form\Extension\Core\Type\TextType;

$this->_form = $this->_app['form.factory']->createBuilder('form', $this->_map())
    ->add('firstName', 'TextType::class', [
        'constraints' => [new Assert\NotBlank()]
    ])
    ->add('lastName', 'TextType::class', [
        'constraints' => [new Assert\NotBlank()]
    ])
    ->add('email', 'TextType::class', [
        'constraints' => [new Assert\Email()]
    ])
    ->getForm();
```

JonathanRamier:
Hello @HeahDude
 
You forgot to remove quote in your example 

```php
use Symfony\Component\Form\Extension\Core\Type\FormType;
use Symfony\Component\Form\Extension\Core\Type\TextType;

$this->_form = $this->_app['form.factory']
[...]

PR #1306, opened 2016-01-15
Address bc changes from Symfony 2.8 to Symfony 3

bestattendance:
Ref: https://github.com/silexphp/Silex/issues/1302

How is pre-Symfony3 documentation handled?  Is that a separate page or should notes on pre-Symfony3 usage be included on this page?

### DIFF
--- a/doc/providers/form.rst
+++ b/doc/providers/form.rst
@@ -69,7 +69,9 @@ Usage
 
 The FormServiceProvider provides a ``form.factory`` service. Here is a usage
 example::
-
+    use Symfony\Component\Form\Extension\Core\Type\FormType;
+    use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+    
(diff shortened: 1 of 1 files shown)


stof:
The Silex master branch currently depends on Symfony 2.7+. The easiest would probably to make it require 2.8+ instead, allowing to use only the new APIs.

dirkluijk:
:+1:  thanks, got these issues as well with the latest Silex version.

ragboyjr:
Is there anything else that needs to be done here? Does this need to be rebased off of master before we can merge in?

fabpot:
As 2.8 is a LTS, I'm going to remove support for 2.7 in Silex 2.0. That could make things easier for everyone.

fabpot:
This one can be merged after #1338

fabpot:
Thank you @bestattendance.

PR #1310, opened 2016-01-25
Allow use of all the RequestMatcher parameters by providing an array

quazardous:
Provider: Silex\Provider\SecurityServiceProvider
Little patch to make full use of the `RequestMatcher` in `$app['security.access_rules']`

```php
    $app['security.access_rules'] = [
        [[
            'host' => 'symfony.com',
            'ips' => '192.168.0.1',
            ...
         ], 'ROLE_ADMIN'],
        ...
    ];
```

### DIFF
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -325,7 +325,7 @@ class SecurityServiceProvider implements ServiceProviderInterface, EventListener
                         'methods' => null,
                         'ips' => null,
                         'attributes' => null,
-                        'schemes' => null
+                        'schemes' => null,
                     ];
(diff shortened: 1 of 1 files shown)


quazardous:
I think it's not me :p

```
  [Composer\Downloader\TransportException]                                     

  The "https://api.github.com/repos/symfony/form/zipball/3bb87faafaa3037dca8a  

  049cc34ab8eeb72d52b8" file could not be downloaded (HTTP/1.1 403 Forbidden)  

                                                                               
```

fabpot:
Thank you @quazardous.

GromNaN:
A test should be added on such feature.

georaldc:
Wouldn't this in theory already be possible by passing a RequestMatcher instance instead as the first argument?

quazardous:
@georaldc you re right, but I thought it was more conf friendly to use a basic array instead of an object... that is the point in testing a string for a simple path request matcher.

georaldc:
@quazardous Makes sense. I can write a test for this if you have no plan to do so.

quazardous:
@georaldc I had no plan... So it would be nice to have one ! Thx a lot.

PR #1315, opened 2016-02-07
Fix env variable in travis.yaml

danez:
SYMFONY_DEPS_VERSION=3 was never evaluated correctly as the before_scripts are comparing with 3.0
The tests were still installing 3.0, as minimumStability=dev, but maybe it is nice fixing that.


fabpot:
Thank you @danez.

Issue #1312, opened 2016-02-02
[Suggestion] use bcrypt by default as encoder


stof:
Currently, the MessageDigestEncoder is used by default (with its default config). I suggest switching to Bcrypt in Silex 2 (This cannot be done in 1.3 as it would break existing apps relying on the default config)

SpacePossum:
:+1:

lschricke:
:+1:

mvanbaak:
:+1:

fabpot:
Anyone willing to work on that one? Should be easy enough.

alcalyn:
`password_hash(): Use of the 'salt' option to password_hash is deprecated`

Yep, that break existing apps, but not only because the hash algorithm changes.
Providing a salt to `password_hash` is now deprecated since early released version 7.0.4 of PHP.

stof:
@alcalyn just return ``null`` as salt for your user, instead of forcing the Symfony class to use your own salt.
and for an existing app, you can still configure the encoder explicitly to keep the old one

PR #1316, opened 2016-02-15
Issue #1312 use BCrypt for password encoding

mvanbaak:


src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -114,7 +114,7 @@ public function register(Container $app)
         };
 
         $app['security.encoder.digest'] = function ($app) {
```
mathroc: the service name is not correct anymore, I think you should have add two others services 

 * `security.encoder.bcrypt` returning the `BCryptPasswordEncoder`
 * `security.default_encoder` returning `$app['security.encoder.bcrypt']`

and use `security.default_encoder` instead of `security.encoder.digest` where necessary

mathroc: bonus, $app['security.encoder.bcrypt.cost'] could be used to configure/override the default encoder cost

mvanbaak: Sorry. But wont have time for it this weekend and next week. 
Thanks for the feedback

mvanbaak:
See PR 1320 for the good implementation

PR #1320, opened 2016-02-27
Use BCrypt by default as encoder #1312

e-moe:
Based on work of @mvanbaak (https://github.com/silexphp/Silex/pull/1316) with applied comments from @mathroc and fixed tests.

### DIFF
--- a/doc/providers/security.rst
+++ b/doc/providers/security.rst
@@ -10,6 +10,8 @@ Parameters
 * **security.hide_user_not_found** (optional): Defines whether to hide user not
   found exception or not. Defaults to ``true``.
 
+* **security.encoder.bcrypt.cost** (optional): Defines BCrypt password encoder cost. Defaults to 13.
+
 Services
 --------
 
@@ -36,9 +38,15 @@ Services
   Request object.
 
 * **security.encoder_factory**: Defines the encoding strategies for user
-  passwords (default to use a digest algorithm for all users).
+  passwords (uses ``security.default_encoder``).
 
-* **security.encoder.digest**: The encoder to use by default for all users.
+* **security.default_encoder**: The encoder to use by default for all users (BCrypt).
+
+* **security.encoder.digest**: Digest password encoder.
+
+* **security.encoder.bcrypt**: BCrypt password encoder.
+
+* **security.encoder.pbkdf2**: Pbkdf2 password encoder.
 
 * **user**: Returns the current user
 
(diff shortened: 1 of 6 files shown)


e-moe:
Merge conflict fixed.

doc/providers/security.rst
```diff
@@ -552,18 +558,16 @@ sample users::
...
 
-    use Symfony\Component\Security\Core\Encoder\MessageDigestPasswordEncoder;
+    use Symfony\Component\Security\Core\Encoder\BCryptPasswordEncoder;
 
     $app['security.encoder.digest'] = function ($app) {
-        // use the sha1 algorithm
-        // don't base64 encode the password
-        // use only 1 iteration
-        return new MessageDigestPasswordEncoder('sha1', false, 1);
+        // use only 10 iterations
```
mathroc: with bcrypt, the number of iterations is actually 2 to the power of `$cost`

mathroc:
looks good to me! :+1:

doc/providers/security.rst
e-moe: thanks for comment @mathroc, updated

mvanbaak:
Thank you for doing my work. If only time did allow me to put my actions where my mouth is :(
(+1)

doc/providers/security.rst
```diff
@@ -552,18 +558,16 @@ sample users::
...
-service::
+By default, Silex uses the ``BCrypt`` algorithm to encode passwords.
+Additionally, the password is encoded multiple times.
+You can change these defaults by overriding the encoding cost
+``security.encoder.bcrypt.cost`` or ``security.default_encoder`` service::
 
-    use Symfony\Component\Security\Core\Encoder\MessageDigestPasswordEncoder;
+    use Symfony\Component\Security\Core\Encoder\BCryptPasswordEncoder;
 
     $app['security.encoder.digest'] = function ($app) {
```
fabpot: should be `bcrypt` here, right?

fabpot:
Would it makes sense to instead use `Pbkdf2PasswordEncoder`? Or at least add a sensible configuration for it by default in the provider?

doc/providers/security.rst
mathroc: indeed, `security.encoder.bcrypt` or maybe `security.default_encoder` (as it could be replaced with anything)

mathroc:
@fabpot I don't have enough knowledge to argue for bcrypt or pbkdf2 but [the php documentation on hash_pbkdf2](php.net/hash_pbkdf2) says : 
>  The PBKDF2 method can be used for hashing passwords for storage. However, it should be noted that password_hash() or crypt() with CRYPT_BLOWFISH are better suited for password storage. 

But yes, a `security.encoder.pbkdf2` could be added as it's also widely used

@e-moe what do you think ?

e-moe:
@mathroc I'm not a security expert either. But I would say let's give users possibility to choose. I will update my pull request

e-moe:
updated to meet comments from @fabpot and @mathroc

doc/providers/security.rst
```diff
@@ -10,6 +10,8 @@ Parameters
 * **security.hide_user_not_found** (optional): Defines whether to hide user not
   found exception or not. Defaults to ``true``.
 
+* **security.encoder.bcrypt.cost** (optional): Defines BCrypt password encoder cost. Defaults to 15.
```
stof: Symfony's SecurityBundle defaults to 13. Shouldn't we be consistent here ?

e-moe: updated

fabpot:
Thank you @e-moe.

PR #1324, opened 2016-03-18
Added test for access rule path array

georaldc:
Related to PR #1310 Also fixed default value for 5th argument passed to RequestMatcher
constructor.


fabpot:
Thank you @georaldc.

PR #1330, opened 2016-04-01
Add the possibility to register forms as services

skalpa:
This allows you to register your form types / form types extensions / form types guessers as Silex services:
```php
    $app['your.type.service'] = function ($app) {
        return new YourServiceFormType();
    };
    $app->extend('form.types', function ($types) use ($app) {
        $types[] = 'your.type.service';
        return $types;
    }));
    ...
    $builder = $app['form.factory']->createBuilder('your.type.service');
```
This is particularly useful for users of Symfony 3.x, now that `FormFactory::createBuilder()` and friends don't allow you to pass an instance of `FormTypeInterface` anymore, preventing you from loading form types that require constructor arguments lazily (without the patch you have to add objects to the `form.types` / `form.type.extensions` / `form.type.guessers` parameters which means every single form of your application will have to get instantiated whenever you use the form component).

The patch contains a new `SilexFormExtension` extension class that handles the dereferencing of services names / lazy loading, new tests, and documentation fixes.

Other quick notes:
* No type checking is performed by the extension, but that's consistent with the way the `FormServiceProvider` works without the patch. Nevertheless, I believe this should be enhanced and will be ready to work on this after the merge if you reckon that would be a good idea.
* I had to change the `DummyFormType` declaration in the tests. You used a class with a `getName()` method with Symfony < 2.8 and one without the method with 2.8+. Now the problem is that the `FormRegistry` from 2.8.x still uses the `getName()` method to store form types, which means it is still needed in some cases. Thus for my tests to pass I had to make sure the class with `getName()` was used with any 2.x version while the one without it was only used with 3.x

### DIFF
--- a/src/Silex/Provider/Form/SilexFormExtension.php
+++ b/src/Silex/Provider/Form/SilexFormExtension.php
@@ -11,23 +11,23 @@
 
 namespace Silex\Provider\Form;
 
-use Silex\Application;
+use Pimple\Container;
 use Symfony\Component\Form\Exception\InvalidArgumentException;
 use Symfony\Component\Form\FormExtensionInterface;
 use Symfony\Component\Form\FormTypeGuesserChain;
 
 class SilexFormExtension implements FormExtensionInterface
 {
-    private $silex;
+    private $app;
     private $types;
     private $typeExtensions;
     private $guessers;
     private $guesserLoaded = false;
     private $guesser;
 
-    public function __construct(Application $silex, array $types, array $typeExtensions, array $guessers)
+    public function __construct(Container $app, array $types, array $typeExtensions, array $guessers)
     {
-        $this->silex = $silex;
+        $this->app = $app;
         $this->setTypes($types);
         $this->setTypeExtensions($typeExtensions);
         $this->setGuessers($guessers);
@@ -39,7 +39,7 @@ class SilexFormExtension implements FormExtensionInterface
             throw new InvalidArgumentException(sprintf('The type "%s" is not the name of a registered form type.', $name));
         }
         if (!is_object($this->types[$name])) {
-            $this->types[$name] = $this->silex[$this->types[$name]];
+            $this->types[$name] = $this->app[$this->types[$name]];
         }
 
         return $this->types[$name];
@@ -69,7 +69,7 @@ class SilexFormExtension implements FormExtensionInterface
                 $guessers = [];
                 foreach ($this->guessers as $guesser) {
                     if (!is_object($guesser)) {
-                        $guesser = $this->silex[$guesser];
+                        $guesser = $this->app[$guesser];
                     }
                     $guessers[] = $guesser;
                 }
@@ -87,7 +87,7 @@ class SilexFormExtension implements FormExtensionInterface
         $this->types = [];
         foreach ($types as $type) {
             if (!is_object($type)) {
-                if (!isset($this->silex[$type])) {
+                if (!isset($this->app[$type])) {
                     throw new InvalidArgumentException(sprintf('Invalid form type. The silex service "%s" does not exist.', $type));
                 }
                 $this->types[$type] = $type;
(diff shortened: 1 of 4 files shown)


HeahDude:
Hi @skalpa, thanks for proposing this!
You can use both `getName()` and `getBlockPrefix()` in `DummyType`:
```php
class DummyType extends AbstractType
{
    // ...
    public function getName()
    {
        return $this->getBlockPrefix();
    }
    // ...
}
```
Besides, I'm not sure of the benefit given that you can already do:
```php
$app['your.type.service'] = function ($app) {
        return new YourServiceFormType();
    };
    $app->extend('form.types', function ($types) use ($app) {
        $types[] = $app['your.type.service'];

        return $types;
    }));
```
Am I missing something ?

HeahDude:
Hmmmm, yes maybe the lazy loading!

skalpa:
@HeahDude Yes, my use case is about lazy loading. The app I am upgrading to SF3 contains several dozens of forms, and instantiating all of them + their dependencies (some of the forms have dependencies injected as constructor arguments) every time I need a single one just feels wrong.

With Symfony 2.7 I used to do something like this (well, to give you the idea... the real code is a bit more complicated considering my controllers don't usually have access to the application instance):
```php
    $form = $app['form.factory']->createBuilder($app['my.something.form']);
```
But this is not possible anymore, so I checked how the `Symfony\Component\Form\Extension\DependencyInjection\DependencyInjectionExtension` worked and did something similar for Silex.

It's mostly useful for form types, but I added support for type extensions and guessers for the sake of consistency.

Also, thanks for your tip about `DummyFormType`, but this was already here and I tried and change as little as possible to existing files. Actually, I intend to propose other changes to this test class later on (if I'm correct one of the tests doesn't actually do anything and should be fixed), but I'll wait until the pull request is merged (or rejected) before I contribute more changes to this file.

HeahDude:
You're right, this is a nice improvement :+1:

But I'm realising that you can already achieve lazy loading by using:
```php
$app['your.type.service'] = YourServiceFormType::class;
```
And the same goes for guessers and form extensions. So we may not need that PR.

HeahDude:
Except if they have dependencies....

skalpa:
Yes, "except if", that's the point ;-)

src/Silex/Provider/Form/SilexFormExtension.php
```diff
@@ -0,0 +1,124 @@
...
+    private $silex;
+    private $types;
+    private $typeExtensions;
+    private $guessers;
+    private $guesserLoaded = false;
+    private $guesser;
+
+    public function __construct(Application $silex, array $types, array $typeExtensions, array $guessers)
+    {
+        $this->silex = $silex;
```
GromNaN: Instances of `Silex\Application` are called `$app` elsewhere. 
The type hint could be done on `Pimple\Container` or even `ArrayAccess` for more reusability.

skalpa:
I changed the Application type hint to Container and renamed the private property to app as advised by @GromNaN
I also removed a piece of code that was dealing with the differences between versions of the Form component now that Silex isn't compatible with versions before 2.8.x
I guess it should be ready to review/merge now.

tests/Silex/Tests/Provider/FormServiceProviderTest.php
```diff
@@ -53,6 +53,51 @@ public function testFormServiceProviderWillLoadTypes()
...
+            return new DummyFormType();
+        };
+        $app->extend('form.types', function ($extensions) {
+            $extensions[] = 'dummy';
+
+            return $extensions;
+        });
+
+        $form = $app['form.factory']
+            ->createBuilder(class_exists('Symfony\Component\Form\Extension\Core\Type\RangeType') ? 'Symfony\Component\Form\Extension\Core\Type\FormType' : 'form', array())
```
fabpot: RangeType always exists as we bumped the min requirement to 2.8 (same everywhere else).

fabpot:
LGTM. Can you rebase to get rid of the merge commit?

skalpa:
@fabpot The branch has been rebased, and all the tests for < 2.8 have been removed from `FormServiceProviderTest` (I actually removed all the tests I could find in the file, not only the ones I had added myself, I hope it's all right).

fabpot:
Thank you @skalpa.

Issue #1331, opened 2016-04-01
Cannot instantiate twig if FormServiceProvider is used without CsrfServiceProvider


skalpa:
On master:
```php
    $app = new Application();
    $app->register(new FormServiceProvider());
    $app->register(new TwigServiceProvider());
    $twig = $app['twig'];
```
This fails because the declaration of `twig.form.renderer` assumes that `csrf.token_manager` exists.

I'll send a fix shortly.

PR #1332, opened 2016-04-02
Fixing TwigServiceProvider when used with FormServiceProvider but without CsrfServiceProvider

skalpa:
Trivial fix for #1331, with test.


ragboyjr:
Can we merge this in?

fabpot:
Thank you @skalpa.

PR #1333, opened 2016-04-06
encode documentation exemple password with password_hash

mathroc:
update documentation hashed password to match the default password encoder because the old hash is not valid with the bcrypt encoder

see #1320


fabpot:
Thank you @mathroc.

Issue #1013, opened 2014-09-12
request is NULL in route conditions


hinzundcode:
I wan't to use route conditions in Silex, but request is NULL, only context is provided.

anasofiapaixao:
I still have this issue, with silex 1.3.5 and symfony/expression-language 3.0.4. This example triggers it:

```php
$app->get('/', function() { return "test"; })
    ->getRoute()->setCondition('request.getPort() === 1234');
```

GromNaN:
@anasofiapaixao You should create a new issue with a runnable code sample that demonstrates the issue.

anasofiapaixao:
Looks like @SpacePossum 's test case in https://github.com/silexphp/Silex/pull/1334 already shows the issue.

PR #1334, opened 2016-04-07
[WIP/RFC] Route with conditional (by expression language) support on 1.x branch

SpacePossum:
proves the problem described here
https://github.com/silexphp/Silex/issues/1013#issuecomment-206963327

this test does pass on master

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "symfony/css-selector": "~2.3|3.0.*",
         "symfony/debug": "~2.3|3.0.*",
         "symfony/dom-crawler": "~2.3|3.0.*",
+        "symfony/expression-language": "~2.3|3.0.*",
         "symfony/finder": "~2.3|3.0.*",
         "symfony/monolog-bridge": "~2.3|3.0.*",
         "symfony/options-resolver": "~2.3|3.0.*",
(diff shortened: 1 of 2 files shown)


GromNaN:
Thanks for the test case.
Your issue is more a feature request than a bug report. The expression engine component is currently not supported by Silex.

SpacePossum:
Thanks @GromNaN, do you know if the support for the expressions in routes on master by design or (more or less) by accident?

GromNaN:
After checking at the actual issue, it seems that the ExpressionLanguage component is automatically available in Routing : https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Routing/Matcher/UrlMatcher.php#L240-L250

But the request object is not accessible because Silex 1.x uses the [`UrlMatcher`](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Routing/Matcher/UrlMatcher.php#L88-L99). The [`RequestMatcher`](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Routing/Matcher/UrlMatcher.php#L104-L113) must be used to have this feature. The great news is that it is supported in Silex 2.x thanks to #971!

GromNaN:
Also, for your use-case, you can user `requireHttps`.

```php
$app->get('/', function () { return 'test'; })->requireHttps();
```

anasofiapaixao:
For anyone coming here via google looking for a workaround, although the `request` object is not available via Silex, the `context` object (of class `Symfony\Component\Routing\RequestContext`) is, which has some information that may be useful (like host, port, etc). For my use case,

``` context.getHost() === 'example.com' ```

did the trick.

fabpot:
see #1334 for proper support in 2.0.

PR #1338, opened 2016-05-03
removed support for Symfony 2.7 in Silex 2.0

fabpot:
As 2.8 is a LTS and compatible with 3.0, it's going to be much easier to maintain both the code and the documentation for Silex 2.0 if we drop support for Symfony 2.7.


composer.json
```diff
@@ -17,31 +17,31 @@
     "require": {
         "php": ">=5.5.9",
         "pimple/pimple": "~3.0",
-        "symfony/event-dispatcher": "~2.7|3.0.*",
-        "symfony/http-foundation": "~2.7|3.0.*",
-        "symfony/http-kernel": "~2.7|3.0.*",
-        "symfony/routing": "~2.7|3.0.*"
+        "symfony/event-dispatcher": "~2.8|3.0.*",
```
stof: why not ``~3.0``, relying on the Symfony BC promise ?

src/Silex/Application/FormTrait.php
```diff
@@ -32,8 +32,7 @@
     public function form($data = null, array $options = array(), $type = null)
     {
         if (null === $type) {
-            // BC with Symfony < 2.8
-            $type = class_exists('Symfony\Component\Form\Extension\Core\Type\RangeType') ? 'Symfony\Component\Form\Extension\Core\Type\FormType' : 'form';
+            $type = 'Symfony\Component\Form\Extension\Core\Type\FormType';
```
stof: this could use the ``::class`` notation

PR #1339, opened 2016-05-03
added route conditions

fabpot:
fixes #1334


SpacePossum:
Thank you @fabpot :+1:

PR #1340, opened 2016-05-03
Clarify purpose of BootableServiceProvider

rbayliss:
`BootableServiceProvider` has an outdated description.  This PR brings the comment into line with the rest of the interfaces in `Silex\Api`


fabpot:
Good catch, thanks @rbayliss.

PR #1341, opened 2016-05-03
Added support for callables in CallbackResolver

### LINKED ISSUE #1327: allow "callable" classes & closure as controller services
what do you think about allowing any [callable](http://php.net/manual/en/language.types.callable.php) as controller services? (or anywhere a callback is expected)

at the moment closure cannot be used as controllers because callback should be objects on which a method is called. but in php there are 2 other kind of callable:

 * objects with an `__invoke` method (can be used like that: 'callable:__invoke')
 * closure

I guess something along those lines could work:

```php
class CallbackResolver
{
    public function isValid($name)
    {
        return is_string($name);
    }

    public function convertCallback($name)
    {
        if (preg_match(static::SERVICE_PATTERN, $name)) {
            list($service, $method) = explode(':', $name, 2);
            if (!isset($this->app[$service])) {
                throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
            }
        
            $callback = array($this->app[$service], $method);
        } else {
            $callback = $this->app[$service];
        }
        if (!is_callable($callback)) {
            throw new \InvalidArgumentException(sprintf('Service "%s" is not callable.', $name));
        }
        return $callback;
    }
```

ragboyjr:
Added the ability for the callback resolver to support any callable instead
of just the object and method callable type.

An alternate implementation to this would be to update the CallbackResolver to use an interface, then keep the old CallbackResolver as is and just add a new implementation that you can change if you'd like. I'd probably prefer to do the interface, but depends on what people think.

This fixes #1327 

### DIFF
--- a/doc/providers/service_controller.rst
+++ b/doc/providers/service_controller.rst
@@ -114,3 +114,29 @@ followed by a single colon (:), followed by the method name.
     };
 
     $app->get('/posts.json', "posts.controller:indexJsonAction");
+
+In addition to using classes for service controllers, you can define any
+callable as a service in the application to be used for a route.
+
+.. code-block:: php
+
+    namespace Demo\Controller;
+
+    use Demo\Repository\PostRepository;
+    use Symfony\Component\HttpFoundation\JsonResponse;
+
+    function postIndexJson(PostRepository $repo) {
+        return function() use ($repo) {
+            return new JsonResponse($repo->findAll());
+        };
+    }
+
+And when defining your route, the code would look like the following:
+
+.. code-block:: php
+
+    $app['posts.controller'] = function($app) {
+        return Demo\Controller\postIndexJson($app['posts.repository']);
+    };
+
+    $app->get('/posts.json', 'posts.controller');
--- a/src/Silex/CallbackResolver.php
+++ b/src/Silex/CallbackResolver.php
@@ -33,7 +33,7 @@ class CallbackResolver
      */
     public function isValid($name)
     {
-        return is_string($name) && preg_match(static::SERVICE_PATTERN, $name);
+        return is_string($name) && (preg_match(static::SERVICE_PATTERN, $name) || isset($this->app[$name]));
     }
 
     /**
@@ -41,19 +41,25 @@ class CallbackResolver
      *
      * @param string $name
      *
-     * @return array A callable array
+     * @return callable A callable value
      *
      * @throws \InvalidArgumentException In case the method does not exist.
      */
     public function convertCallback($name)
     {
-        list($service, $method) = explode(':', $name, 2);
-
-        if (!isset($this->app[$service])) {
-            throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
+        if (preg_match(static::SERVICE_PATTERN, $name)) {
+            list($service, $method) = explode(':', $name, 2);
+            $callback = array($this->app[$service], $method);
+        } else {
+            $service = $name;
+            $callback = $this->app[$name];
         }
 
-        return array($this->app[$service], $method);
+        if (!is_callable($callback)) {
+            throw new \InvalidArgumentException(sprintf('Service "%s" is not callable.', $service));
+        }
+
+        return $callback;
     }
 
     /**
@@ -61,7 +67,7 @@ class CallbackResolver
      *
      * @param string $name
      *
-     * @return array A callable array
+     * @return string|callable A callable value or the string passed in
      *
      * @throws \InvalidArgumentException In case the method does not exist.
      */
--- a/tests/Silex/Tests/CallbackResolverTest.php
+++ b/tests/Silex/Tests/CallbackResolverTest.php
@@ -27,27 +27,55 @@ class CallbackResolverTest extends \PHPUnit_Framework_Testcase
 
     public function testShouldResolveCallback()
     {
-        $this->app['some_service'] = function () { return new \stdClass(); };
+        $callable = function () {};
+        $this->app['some_service'] = function () { return new \ArrayObject(); };
+        $this->app['callable_service'] = function () use ($callable) {
+            return $callable;
+        };
 
         $this->assertTrue($this->resolver->isValid('some_service:methodName'));
+        $this->assertTrue($this->resolver->isValid('callable_service'));
         $this->assertEquals(
-            array($this->app['some_service'], 'methodName'),
-            $this->resolver->convertCallback('some_service:methodName')
+            array($this->app['some_service'], 'append'),
+            $this->resolver->convertCallback('some_service:append')
         );
+        $this->assertSame($callable, $this->resolver->convertCallback('callable_service'));
     }
 
-    public function testNonStringsAreNotValid()
+    /**
+     * @dataProvider nonStringsAreNotValidProvider
+     */
+    public function testNonStringsAreNotValid($name)
     {
-        $this->assertFalse($this->resolver->isValid(null));
-        $this->assertFalse($this->resolver->isValid('some_service::methodName'));
+        $this->assertFalse($this->resolver->isValid($name));
+    }
+
+    public function nonStringsAreNotValidProvider()
+    {
+        return array(
+            array(null),
+            array('some_service::methodName'),
+            array('missing_service'),
+        );
     }
 
     /**
      * @expectedException          \InvalidArgumentException
-     * @expectedExceptionMessage   Service "some_service" does not exist.
+     * @expectedExceptionMessageRegExp  /Service "[a-z_]+" is not callable./
+     * @dataProvider shouldThrowAnExceptionIfServiceIsNotCallableProvider
      */
-    public function testShouldThrowAnExceptionIfServiceIsMissing()
+    public function testShouldThrowAnExceptionIfServiceIsNotCallable($name)
     {
-        $this->resolver->convertCallback('some_service:methodName');
+        $this->app['non_callable_obj'] = function () { return new \stdClass(); };
+        $this->app['non_callable'] = function () { return array(); };
+        $this->resolver->convertCallback($name);
+    }
+
+    public function shouldThrowAnExceptionIfServiceIsNotCallableProvider()
+    {
+        return array(
+            array('non_callable_obj:methodA'),
+            array('non_callable'),
+        );
     }
 }


src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,64 @@ public function __construct(Container $app)
      */
     public function isValid($name)
     {
-        return is_string($name) && preg_match(static::SERVICE_PATTERN, $name);
+        return is_string($name) && (
+            preg_match(static::SERVICE_PATTERN, $name) ||
+            isset($this->app[$name])
+        );
```
fabpot: That should be on one line.

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,64 @@ public function __construct(Container $app)
...
+    }
+
+    /**
+     * Throws an InvalidArgumentException for the service not existing.
+     *
+     * @param string $name
+     *
+     * @throws \InvalidArgumentException
+     */
+    protected function throwServiceNotExists($service)
```
fabpot: Creating a `private` method sounds better to me (should be moved at the end of the file).

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,64 @@ public function __construct(Container $app)
...
+        return is_string($name) && (
+            preg_match(static::SERVICE_PATTERN, $name) ||
+            isset($this->app[$name])
+        );
+    }
+
+    /**
+     * Throws an InvalidArgumentException for the service not existing.
+     *
+     * @param string $name
```
fabpot: $service, not $name but anyway, the whole docblock should be removed as we don't use them for private methods (except when it is really needed, which is not the case here).

ragboyjr:
@fabpot, those are fixed.

src/Silex/CallbackResolver.php
```diff
@@ -33,40 +33,59 @@ public function __construct(Container $app)
      */
     public function isValid($name)
     {
-        return is_string($name) && preg_match(static::SERVICE_PATTERN, $name);
+        return is_string($name) && (preg_match(static::SERVICE_PATTERN, $name) || isset($this->app[$name]));
```
mathroc: I'm not sure when isValid() is called, does this means services have to be created before they can be used ?
why not keep just is_string() here and rely on convertCallback to throw exception when needed?

ragboyjr: @mathroc 

Because it's used here. https://github.com/silexphp/Silex/blob/1.3/src/Silex/ServiceControllerResolver.php#L46 

I thought about just doing that as well, but I wanted to keep it as BC as possible.

src/Silex/CallbackResolver.php
```diff
@@ -33,40 +33,59 @@ public function __construct(Container $app)
...
      */
     public function resolveCallback($name)
     {
         return $this->isValid($name) ? $this->convertCallback($name) : $name;
     }
+
+    private function throwServiceNotExists($service)
+    {
+        throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
+    }
```
fabpot: I didn't had a look at the code, just the coding standard but what about inlining this method? I think having the same exception duplicated twice is fine.

ragboyjr: It's your project :), if you want it twice, I'll paste it twice.

ragboyjr: @fabpot, it's updated now.

src/Silex/CallbackResolver.php
```diff
@@ -33,40 +33,59 @@ public function __construct(Container $app)
...
     {
-        list($service, $method) = explode(':', $name, 2);
+        if (preg_match(static::SERVICE_PATTERN, $name)) {
+            list($service, $method) = explode(':', $name, 2);
 
-        if (!isset($this->app[$service])) {
-            throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
-        }
+            if (!isset($this->app[$service])) {
+                $this->throwServiceNotExists($service);
```
mathroc: is it really useful to have this? Pimple already throw an exception if the key does not exists

alternatively, the whole if else could be wrapped in a `try catch(\InvalidArgumentException)` and rethrow another with a more explicit message if needed

```php
    public function convertCallback($name)
    {
        try {
            if (preg_match(static::SERVICE_PATTERN, $name)) {
                list($service, $method) = explode(':', $name, 2);
                return array($this->app[$service], $method);
            } else {
                $callback = $this->app[$name];

                if (!is_callable($callback)) {
                    throw new \InvalidArgumentException(sprintf('Service "%s" is not callable.', $name));
                }

                return $callback;
            }
    }
    catch (\InvalidArgumentException $e) {
        throw new \InvalidArgumentException(sprintf('Service "%s" is not a valid callback.', $service), 0, $e);
    }
```

ragboyjr: Good point about Pimple already throwing an exception. I was just copying the design of the code that was there before. I'm not a huge fan of rethrowing, but wouldn't mind removing the checks to see if the service actually exists and let pimple throw an error message.

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,49 @@ public function __construct(Container $app)
...
 
-        if (!isset($this->app[$service])) {
-            throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
-        }
+            if (!isset($this->app[$service])) {
+                throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
+            }
+
+            return array($this->app[$service], $method);
+        } else {
```
HeahDude: The `else` is not needed because of the `return` at the line before.

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,49 @@ public function __construct(Container $app)
...
     public function convertCallback($name)
     {
-        list($service, $method) = explode(':', $name, 2);
+        if (preg_match(static::SERVICE_PATTERN, $name)) {
+            list($service, $method) = explode(':', $name, 2);
 
-        if (!isset($this->app[$service])) {
-            throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
-        }
+            if (!isset($this->app[$service])) {
```
HeahDude: This check is done in `isValid()`, right? Why not test if the method is callable instead?

tests/Silex/Tests/CallbackResolverTest.php
```diff
@@ -28,26 +28,58 @@ public function setup()
     public function testShouldResolveCallback()
     {
         $this->app['some_service'] = function () { return new \stdClass(); };
+        $this->app['callable_service'] = function () { return function () {}; };
 
         $this->assertTrue($this->resolver->isValid('some_service:methodName'));
+        $this->assertTrue($this->resolver->isValid('callable_service'));
```
HeahDude: I think you could add another test for this.

src/Silex/CallbackResolver.php
ragboyjr: 1. This is a public method, so it technically can be called with isValid being called.
2. To test if is_callable, you need to fetch the service from the container which may be a lot of work. In isValid, I don't think we should load the service.

src/Silex/CallbackResolver.php
ragboyjr: Good catch

src/Silex/CallbackResolver.php
HeahDude: 1. You can use `isValid` at the very beginning of this method to avoid the duplication below of this test.
2. At this point the service will surely be called, I guess we could test the array here as you do below for closures.

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,49 @@ public function __construct(Container $app)
...
+                throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
+            }
+
+            return array($this->app[$service], $method);
+        } else {
+            if (!isset($this->app[$name])) {
```
stof: I would remove this check, and just let Pimple throw the exception itself when accessing ``$this->app[$name]`` below

ragboyjr: @stof, ya, I think I'm cool with that. Anyone else have any objects to that?

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,49 @@ public function __construct(Container $app)
...
-        if (!isset($this->app[$service])) {
-            throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
+            if (!isset($this->app[$service])) {
+                throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
+            }
+
+            return array($this->app[$service], $method);
+        }
+
+        if (!isset($this->app[$name])) {
```
ragboyjr: @HeahDude, ok, so convertCallback assumes that the name is valid. So doing the `isset` test on line 60 is redundant, so I will remove.

However, the check at line 53 is different because we aren't testing if `$name` is in the container, but if the service part is in. The `$service` would be the first part of the `$name` var.

HeahDude: I reword my previous comment: since you call the service line 57 which would throw a pimple exception, why not change the test line 53 to `if (!is_callable(array($this->app[$service], $method)))`?

ragboyjr: Ok, my original reason was for BC purposes to make this change as little as possible. But, with all of these other changes, it'd be better to do a check for is callable on the object based callables.

doc/providers/service_controller.rst
```diff
@@ -114,3 +114,29 @@ followed by a single colon (:), followed by the method name.
     };
 
     $app->get('/posts.json', "posts.controller:indexJsonAction");
+
+In addition to using classes for service controllers, you can define any
```
HeahDude: "In addition OF using"?

doc/providers/service_controller.rst
```diff
@@ -114,3 +114,29 @@ followed by a single colon (:), followed by the method name.
...
+callable as a service in the application to be used for a route.
```
HeahDude: "to be used as controller"?

doc/providers/service_controller.rst
```diff
@@ -114,3 +114,29 @@ followed by a single colon (:), followed by the method name.
...
+    use Demo\Repository\PostRepository;
+    use Symfony\Component\HttpFoundation\JsonResponse;
+
+    function postIndexJson(PostRepository $repo) {
+        return function() use ($repo) {
+            return new JsonResponse($repo->findAll());
+        };
+    }
+
+And in your route definition would look like the following code.
```
HeahDude: "And when defining your route, the code would look like the following:"?

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,45 @@ public function __construct(Container $app)
...
+        return is_string($name) && (preg_match(static::SERVICE_PATTERN, $name) || isset($this->app[$name]));
     }
 
     /**
      * Returns a callable given its string representation.
      *
      * @param string $name
      *
-     * @return array A callable array
+     * @return callable A callable value
```
HeahDude: I think you can just keep `@return callable`.

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,45 @@ public function __construct(Container $app)
...
+        return $callback;
     }
 
     /**
      * Returns a callable given its string representation if it is a valid service method.
      *
      * @param string $name
      *
-     * @return array A callable array
+     * @return string|callable A callable value or the string passed in
```
HeahDude: Isn't a returned string callable? If so it could just be `@return callable`.

src/Silex/CallbackResolver.php
```diff
@@ -33,35 +33,41 @@ public function __construct(Container $app)
...
 
-        if (!isset($this->app[$service])) {
-            throw new \InvalidArgumentException(sprintf('Service "%s" does not exist.', $service));
+        if (!is_callable($callback)) {
+            throw new \InvalidArgumentException(sprintf('Service "%s" is not callable.', $service));
         }
 
-        return array($this->app[$service], $method);
+        return $callback;
     }
```
HeahDude: Good work, this method really looks nice now :)

doc/providers/service_controller.rst
ragboyjr: http://forum.wordreference.com/threads/in-addition-of-in-additon-to.1358265/

tests/Silex/Tests/CallbackResolverTest.php
```diff
@@ -27,27 +27,54 @@ public function setup()
...
-        $this->app['some_service'] = function () { return new \stdClass(); };
+        $callable = function () {};
+        $this->app['some_service'] = function () { return new \ArrayObject(); };
+        $this->app['callable_service'] = function () use ($callable) {
+            return $callable;
+        };
 
         $this->assertTrue($this->resolver->isValid('some_service:methodName'));
+        $this->assertTrue($this->resolver->isValid('callable_service'));
         $this->assertEquals(
```
HeahDude: Can't you use `assertSame`?

src/Silex/CallbackResolver.php
ragboyjr: No, because it might not be a callable string. It's either a callable or the string passed in. The string passed in could be anything.

tests/Silex/Tests/CallbackResolverTest.php
ragboyjr: yup

src/Silex/CallbackResolver.php
HeahDude: I think I'm confused by the previous `@return array A callable array` since any invalid value would be returned as is. But I guess the point is to only return a callable, right?
If I understand correctly we should either write `@return callable` or `@return mixed`.

ragboyjr: I think the old `array A callable array` was wrong because it obviously returned either an array, or a string according to the docblock type hints (php wouldn't validate a string was passed in anyways).

And https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.param.pkg.html, using the pipe symbol to delimit the return types is valid phpdoc.

ragboyjr:
Anything else I'm missing on this one?

mathroc:
LGTM !

fabpot:
Thank you @ragboyjr.

PR #1345, opened 2016-05-08
Allowing callables for mount

ragboyjr:
- Added the ability to pass callables into the controller collection
    mount method.
- Updated documentation to show usage of new callables in mount and
    recursive mounting.
- Added appropriate tests

This addresses #1262 and #1290 

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -437,8 +437,8 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
     /**
      * Mounts controllers under the given route prefix.
      *
-     * @param string                                           $prefix      The route prefix
-     * @param ControllerCollection|ControllerProviderInterface $controllers A ControllerCollection or a ControllerProviderInterface instance
+     * @param string                                                    $prefix      The route prefix
+     * @param ControllerCollection|callable|ControllerProviderInterface $controllers A ControllerCollection, a callable, or a ControllerProviderInterface instance
      *
      * @return Application
      *
@@ -454,8 +454,8 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
             }
 
             $controllers = $connectedControllers;
-        } elseif (!$controllers instanceof ControllerCollection) {
-            throw new \LogicException('The "mount" method takes either a "ControllerCollection" or a "ControllerProviderInterface" instance.');
+        } elseif (!$controllers instanceof ControllerCollection && !is_callable($controllers)) {
+            throw new \LogicException('The "mount" method takes either a "ControllerCollection" instance, "ControllerProviderInterface" instance, or a callable.');
         }
 
         $this['controllers']->mount($prefix, $controllers);
--- a/src/Silex/ControllerCollection.php
+++ b/src/Silex/ControllerCollection.php
@@ -44,11 +44,13 @@ class ControllerCollection
     protected $defaultController;
     protected $prefix;
     protected $routesFactory;
+    protected $controllersFactory;
 
-    public function __construct(Route $defaultRoute, $routesFactory = null)
+    public function __construct(Route $defaultRoute, RouteCollection $routesFactory = null, $controllersFactory = null)
     {
         $this->defaultRoute = $defaultRoute;
         $this->routesFactory = $routesFactory;
+        $this->controllersFactory = $controllersFactory;
         $this->defaultController = function (Request $request) {
             throw new \LogicException(sprintf('The "%s" route must have code to run when it matches.', $request->attributes->get('_route')));
         };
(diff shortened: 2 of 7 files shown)


ragboyjr:
I had to add a `$controllersFactory` parameter to the ControllerCollection because the new mount method can act as a factory method for creating ControllerCollections. I started out with only trying to do this from the `Application` class, however, the nesting/recursive routing wouldn't work, so I took it out and moved it into the controller collection. Having the ControllerCollections act like a factory might seem like it's going against Separation of concerns principle, but I couldn't see a better way to implement this feature.

```php
$app->mount('/admin', function($admin) {
    $admin->mount('/blog', function($blog) {
        $blog->get('/);
    });
});
```

The above is only possible if the controller collection `mount` method is able to create else we'd have to do something like the following.

```php
$app->mount('/admin', function($admin) use ($app) {
    // either we don't allow recursive routing, which might seem inconsistent for users
    // or we do something like this
    $app->mount($admin, '/blog', function($blog) {
        $blog->get('/');
    });
});
```

The syntax of the second isn't as clean or intuitive as the first; however the Silex code structure is a little cleaner in my opinion, but I can go both ways on this.

ragboyjr:
Any word?

src/Silex/Application.php
```diff
@@ -454,8 +454,8 @@ public function mount($prefix, $controllers)
             }
 
             $controllers = $connectedControllers;
-        } elseif (!$controllers instanceof ControllerCollection) {
-            throw new \LogicException('The "mount" method takes either a "ControllerCollection" or a "ControllerProviderInterface" instance.');
+        } elseif (!$controllers instanceof ControllerCollection && !is_callable($controllers)) {
+            throw new \LogicException('The "mount" method takes either a "ControllerCollection" instance, "ControllerProviderInterface" instance, or a callable.');
```
SpacePossum: `"ControllerProviderInterface" instance` I get what you try to state but it is impossible to have an instance of an interface.
How about
```php
sprintf('The "mount" method takes either a "ControllerCollection", "ControllerProviderInterface" implementation, or `callable`. Got: "%s".', is_object($controllers) ? get_class($controllers) : gettype($controllers)));
```

ragboyjr: @SpacePossum, I was trying to keep consistent with old error, and instances implement interfaces. I prefer the way the error currently is, but I can change if that's consensus. Either way, this error message is a minor part of the whole PR, I'm more interested in seeing if this design change should be accepted or not, then we can work on the little things.

src/Silex/ControllerCollection.php
```diff
@@ -57,11 +59,21 @@ public function __construct(Route $defaultRoute, $routesFactory = null)
...
+     *
+     * @throws \LogicException
      */
-    public function mount($prefix, ControllerCollection $controllers)
+    public function mount($prefix, $controllers)
     {
+        if (is_callable($controllers)) {
+            $collection = $this->controllersFactory ? call_user_func($this->controllersFactory) : new static(new Route(), new RouteCollection());
+            call_user_func($controllers, $collection);
+            $controllers = $collection;
```
fabpot: Why not moving this part of the code to the main `Application::mount()` method?

ragboyjr: That's what I started to do, but then I realized that if we only did it in the application mount then the following code would not work:

```php
$app->mount('/admin', function($admin) {
    $admin->mount('/blog', function($blog) {
        $blog->get('/');
    });
});
```

instead, you'd have to do

```php
$app->mount('/admin', function($admin) use ($app) {
    // either we don't allow recursive routing, which might seem inconsistent for users
    // or we do something like this
    $app->mount($admin, '/blog', function($blog) {
        $blog->get('/');
    });
});
```

ragboyjr: Personally, I like the syntax of the first option, but the silex src code is a little less maintainable :(. I'm ok with the second option. Definitely is a much cleaner patch into the codebase; however, we'd have to clearly let users know that the mount with a function **only** works from the `Silex\Application`

ragboyjr: @fabpot, I just looked at how the [Slim Framework](http://www.slimframework.com/docs/objects/router.html#route-groups) does it. There implementation is that great either, so I think we could get away with just doing the update in the application mount and leave the ControllerCollection class alone. But document it clearly..

fabpot: I see. Fair enough. Let's do it like this.

ragboyjr: :), @fabpot, revert back to just adding the code in the application mount, or keep it as it currently is?

fabpot:
Thank you @ragboyjr.

ragboyjr:
awesome!! thanks!

src/Silex/ControllerCollection.php
SpacePossum: @ragboyjr sorry to get back to this, but I'm kind of confused about the final implementation.
Little example about what I try to do (always easier then typing lots of text ;) )

```php
// foo/module/events/test
class FooEventsProvider implements \Silex\Api\ControllerProviderInterface
{
    public function connect(\Silex\Application $app)
    {
        $controllers = $app['controllers_factory'];
        $controllers->get('/test', function($id){ return "events ".$id;});
        return $controllers;
    }
}

//foo/module/bars/test
class FooBarsProvider implements \Silex\Api\ControllerProviderInterface
{
    public function connect(\Silex\Application $app)
    {
        $controllers = $app['controllers_factory'];
        $controllers->get('/test', function($id){ return "bars ".$id;});
        return $controllers;
    }
}

$app->mount('/foo/{id}', function ($foo) {
    $foo->mount('/events', new FooEventsProvider());
    $foo->mount('/bars', new FooBarsProvider());
});
```

This won't work because the ControllerCollection cannot handle ControllerProviderInterface instances (for that it needs a ref. to the Application). I think this inline with your first sample which is not implemented (all good).
So switch to the second sample I tried to setup the mounting like this;

```php
$app->mount('/foo/{id}', function ($foo) use ($app) {
    $app->mount($foo, '/events', new FooEventsProvider());
    $app->mount($foo, '/bars', new FooBarsProvider());
});
```

This doesn't work because the Application::mount expects two parameters and not 3. 

Any suggestions how to do this?

mantis: I tried to do the same as SpacePossum about 2 weeks ago - and hit same 'problem'. My expectation was one could do this to save loading providers i.e. i'm assuming FooBarsProvider would only get loaded by PHP if the path matched /foo/{id}

PR #1353, opened 2016-05-16
[WIP] Added support for an EmptyPathRoute

### LINKED ISSUE #149: Routing does not handle trailing slashes consistently
This may be merely something to add to the docs.

Using a simple application such as the following:

```
<?php

require_once __DIR__.'/../silex.phar';
use Symfony\Component\HttpFoundation\Response;

$app = new Silex\Application();

$app->get('/', function () use ($app) {
      return new Response('Home root');
});
$app->get('/hello', function () use ($app) {
      return new Response('No trailing slash');
});
$app->get('/hello/', function () use ($app) {
      return new Response('With trailing slash');
});

$app->run();
```

this is the observed behavior:
- http://app/ -> caputured by 'Home' route
- http://app/hello -> caputured by 'No trailing slash' route
- http://app/hello/ -> caputured by 'With trailing slash' route

However expected behaviour would be that the last route is inaccessible:
- http://app/ -> caputured by 'Home' route
- http://app/hello -> caputured by 'No trailing slash' route
- http://app/hello/ -> caputured by 'No trailing slash' route

If the last route is not defined, the result of requesting /hello/ is "Sorry, the page you are looking for could not be found." - but you'd expect it to match the routefor /hello.

This is testing on an arch linux machine running nginx - though that shouldn't really be much of a factor.

ragboyjr:
This would potentially resolve #149 allowing empty paths.

Usage would be like so:

``` php
<?php

// php -S localhost:8000 app.php

require_once __DIR__ . '/vendor/autoload.php';

$app = new Silex\Application();
$app['route_class'] = Silex\Route\EmptyPathRoute::class;


$app->mount('/blog', function($blog) {
    $blog->get('', function() {
        return 'blog main';
    });
    $blog->get('/path', function() {
        return 'blog path';
    });
});

$app->run();
```

I haven't added documentation on this feature, and I'll have to remove the app.php file that I added (it's there for demo/testing purposes). 

There are a lot of directions this can go.
- This can be a third party library that people integrate with (so not apart of the silex core). 
- Or we could keep it like the example shows right now
- Or we can add a parameter in the app named `allow_empty_paths` and that will trigger the route class to use EmptyPathRoute or the normal Route
- Or we can make the EmptyPathRoute the standard Route (meaning, we apply the changes in empty path route to the `Silex\Route` class)
- Or We could only allow empty path routes on nested controllers, and transparently change the base route for nested routes.

### DIFF
--- a/app.php
+++ b/app.php
@@ -1,0 +1,19 @@
+<?php
+
+// php -S localhost:8000 app.php
+
+require_once __DIR__.'/vendor/autoload.php';
+
+$app = new Silex\Application();
+$app['route_class'] = Silex\Route\EmptyPathRoute::class;
+
+$app->mount('/blog', function ($blog) {
+    $blog->get('', function () {
+        return 'blog main';
+    });
+    $blog->get('/path', function () {
+        return 'blog path';
+    });
+});
+
+$app->run();
--- a/src/Silex/Route/EmptyPathRoute.php
+++ b/src/Silex/Route/EmptyPathRoute.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the Silex framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Silex\Route;
+
+use Silex\Route as BaseRoute;
+
+/**
+ * Allows Empty Paths in routes. Useful for mounting.
+ *
+ * @author RJ Garcia <rj@bighead.net>
+ */
+class EmptyPathRoute extends BaseRoute
+{
+    private $pathIsEmpty = false;
+
+    public function getPath()
+    {
+        if (!$this->pathIsEmpty) {
+            return parent::getPath();
+        }
+
+        return '';
+    }
+
+    public function setPath($path)
+    {
+        $this->pathIsEmpty = $path === '';
+
+        return parent::setPath($path);
+    }
+
+    public function serialize()
+    {
+        $data = unserialize(parent::serialize());
+        $data['path_is_empty'] = $this->pathIsEmpty;
+
+        return serialize($data);
+    }
+
+    public function unserialize($serialized)
+    {
+        $data = unserialize($serialized);
+        $this->pathIsEmpty = $data['path_is_empty'];
+
+        parent::unserialize($serialized);
+    }
+}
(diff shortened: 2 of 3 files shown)


ragboyjr:
@AD7six, @alexkappa any comments?

AD7six:
@ragboyjr I haven't used silex for a number of years :)

GromNaN:
I prefers the 1st option. You can share this custom `Route` class in an external party library.
I created several apps using Silex and never had to implement such behavior.

When I mount a `ControllerCollection`, it is usually to split a sub-section of the site. And the root URL of this section usually have an ending slash: `http://example.com/forum/`. Thanks to the `RedirectableUrlMatcher`, requests to `http://example.com/forum` are automatically redirected.

ragboyjr:
@GromNaN Ya, to be honest, I haven't needed it either, but if it was there, I'd probably use it.

alexkappa:
@ragboyjr perhaps the first approach is a good way to go if more people dont agree with this.

Like @GromNaN said some people may never experience this problem, and its fairly easy to work around it as you demonstrated.

Personally I'd like to see this being part of silex core and never have to think about the framework prefixing my routes with leading slashes. But for this to be done a discussion needs to be opened up with one of the maintainers.

ragboyjr:
@alexkappa agreed

GDmac:
Can you test this? This seems to make the problem the other way around now:
``/blog``  blog main
``/blog/`` (No route found for "GET /blog/")

and if you add ``$blog->get('/', function() { .. }`` it's back to square 1, 
redirecting 301 ``/blog`` to``/blog/``

ragboyjr:
@GDmac, not necessarily. Because for RESTFul API's, you don't want the extra `/` added to your routes. 

And with EmptyPathRoute, you have the ability to define either route for `/blog` or `/blog/` via.

```php
$blog->get('', function() {}); // goes to /blog
$blog->get('/', function() {}); // goes to /blog/
```

Without empty path route, it's impossible to get the first route.

GDmac:
Thanks, this works nice on mounted controller(collection). +1
While this will allow access to ``/blog``, there's still the issue 
with 301 redirects on ``/blog/12`` to ``/blog/12/`` (trailing slash) from #149

jlayt:
As someone using Silex for RESTful APIs and frustrated at the mounted controller collection issue #149 , I'd be in favour of this being a config option in core, defaulting to off for backwards compatibility.

seblucas:
Thanks, this was the only way I found to allow accessing to `/databases` (without the trailing slash) without having a 301 redirect with this configuration : 

```
$this->mount('/databases', function ($databases) {
            $databases->get('', 'controller.DatabaseList:getAll');
            ........
});
```

fabpot:

PR #1354, opened 2016-05-16
Adding a flag to disable the Monolog error handler

skalpa:
The monolog error handler can now be disabled by setting the value of the new `monolog.use_error_handler` parameter to false (since it was added in  a53ce5902983493c53eea61f6cf1e304c0aa6409 it was always enabled).

The error handler is disabled by default when the application `debug` parameter is set to true, to ensure it won't silence exceptions and/or errors. 

Without the flag, an uncaught exception thrown during boot on `master` will just result in a white page, even when using `debug` and `error_reporting` (because the error handler calls `exit` after logging exceptions).

### DIFF
--- a/doc/providers/monolog.rst
+++ b/doc/providers/monolog.rst
@@ -31,6 +31,14 @@ Parameters
 * **monolog.exception.logger_filter** (optional): An anonymous function that
   filters which exceptions should be logged.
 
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @skalpa.

PR #1355, opened 2016-05-16
Fixed var name in organizing controllers

ragboyjr:
Updated the var name in the organizing controllers documenation
in order be consistent with the variable naming throughout the
code sample.


fabpot:
Thank you @ragboyjr.

PR #1360, opened 2016-05-18
Registering the silex form extension last to ensure user specified ex…

### LINKED ISSUE #1358: Issue or new behaviour on form extension
Since the PR https://github.com/silexphp/Silex/pull/1330 it's no more possible to disable csrf through a extension form because the order to load the extension is diffrent. 

I have a form extension 
```
class DisableCSRFExtension extends AbstractTypeExtension
{
    public function configureOptions(OptionsResolver $resolver)
    {
        $resolver->setDefaults(array(
            'csrf_protection' => false,
        ));
    }

    public function getExtendedType()
    {
        return method_exists(AbstractType::class, 'getBlockPrefix')
            ? FormType::class
            : 'form';
    }
}
```

I load this extension with
```
        $app['form.type.extensions'] = $app->extend('form.type.extensions', function ($extensions) {
            $extensions[] = new DisableCSRFExtension();
            return $extensions;
        });
```

However in [FormServiceProvider](https://github.com/silexphp/Silex/blob/master/src/Silex/Provider/FormServiceProvider.php#L70) we have 
```
        $app['form.extensions'] = function ($app) {
            $extensions = array(
                $app['form.extension.silex'],
                new HttpFoundationExtension(),
            );
            if (isset($app['csrf.token_manager'])) {
                $extensions[] = $app['form.extension.csrf'];
            }
            if (isset($app['validator'])) {
                $extensions[] = new FormValidatorExtension($app['validator']);
            }
            return $extensions;
        };
```
The form extension csrf is loaded after the custom extensions and my custom extension does not work.
I can fix the issue in overriding `$app['form.extensions']` but I think the default behaviour is not correct

skalpa:
…tensions can modify the behavior of the CSRF and Validator extensions.

Fixes #1358

### DIFF
--- a/src/Silex/Provider/FormServiceProvider.php
+++ b/src/Silex/Provider/FormServiceProvider.php
@@ -69,7 +69,6 @@ class FormServiceProvider implements ServiceProviderInterface
 
         $app['form.extensions'] = function ($app) {
             $extensions = array(
-                $app['form.extension.silex'],
                 new HttpFoundationExtension(),
             );
 
@@ -80,6 +79,7 @@ class FormServiceProvider implements ServiceProviderInterface
             if (isset($app['validator'])) {
                 $extensions[] = new FormValidatorExtension($app['validator']);
             }
+            $extensions[] = $app['form.extension.silex'];
 
             return $extensions;
         };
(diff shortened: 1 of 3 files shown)


fabpot:
Thank you @skalpa.

PR #1362, opened 2016-05-18
Added monolog.formatter.output and monolog.formatter.dateformat param…

grunch:
…eters to format the line


fabpot:
Thank you @grunch.

fabpot:
I've merged and then reverted as I find it easier to just override the service itself.

PR #1366, opened 2016-05-19
Fix installation instruction in readme

jdreesen:
Silex 2 has been released yesterday so `@dev` should be removed from the composer version string.


fabpot:
Thank you @jdreesen.

PR #1381, opened 2016-06-01
SessionServiceProvider: Remove unused property

hansott:


GromNaN:
Well spotted. A relic from Silex 1.

fabpot:
Good catch, thanks @hansott.

PR #1382, opened 2016-06-02
Included extend() explanation in services documentation chapter

hvt:
Together with the Twig add global and filter example.

### DIFF
--- a/doc/services.rst
+++ b/doc/services.rst
@@ -176,7 +176,7 @@ in the second argument.
 
     Be sure to return the modified service in the callable.
 
-You can use this pattern to add functionality to :doc:Monolog <providers/twig> for
+You can use this pattern to add functionality to :doc:Twig <providers/twig> for
 example::
 
     $app->extend('twig', function($twig, $app) {


doc/services.rst
```diff
@@ -161,6 +161,31 @@ using the ``protect`` method::
 
 Note that protected closures do not get access to the container.
 
+Modify services after definition
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes you want to alter a service after its definition. Pimple facilitates
+this by ``extend``-ing the already defined service.
```
stof: I would just say ``by extending``, without using a code block in the middle

doc/services.rst
```diff
@@ -161,6 +161,31 @@ using the ``protect`` method::
...
+
+First argument of the ``extend`` method is the name of the service you want to
+modify. Second argument is a callable. This callable is executed with the service
+you want to alter as its first argument, the service container self is provided
```
stof: should be ``itself``, not ``self``

GromNaN:
This is a feature of Pimple. You improve the Pimple documentation and link to it : http://pimple.sensiolabs.org/#modifying-services-after-definition

stof:
Linking to the Pimple documentation is indeed a much better idea.

doc/services.rst
hvt: Tried avoiding using `it` twice in consecutive sentences, but wil fix.

hvt:
So why are there descriptions of `factory` and `protect`? These are both also Pimple features.

Besides, the `extend` example is already on the Twig provider page. Some explanation of what `extend` is and does, seems appropriate to me.

fabpot:
Thank you @hvt.

PR #1385, opened 2016-06-09
Fix deprecations under Symfony 3.1

romainneutron:


src/Silex/ArgumentResolver.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function setApplication(Application $app)
+    {
+        $this->app = $app;
+    }
+
+    public function getArguments(Request $request, $controller)
+    {
+        if (null === $this->app) {
+            throw new \LogicException('You should inject Silex Application prior to call ArgumentResolver::getArguments');
```
GromNaN: Since the app is a required dependency and is injected just after the construct, why not make it part of the constructor signature ?

src/Silex/ArgumentResolver.php
```diff
@@ -0,0 +1,62 @@
...
+ * @author Romain Neutron <imprec@gmail.com>
+ */
+class ArgumentResolver implements ArgumentResolverInterface
+{
+    private $ar;
+    private $app;
+
+    public function __construct(ArgumentResolverInterface $ar)
+    {
+        $this->ar = $ar;
```
GromNaN: Short variable names are bad for readability. What about naming it `$argumentResolver`?

tests/Silex/Tests/ControllerResolverTest.php
```diff
@@ -24,6 +25,10 @@ class ControllerResolverTest extends \PHPUnit_Framework_TestCase
 {
     public function testGetArguments()
     {
+        if (Kernel::VERSION_ID >= 30100) {
+            $this->markTestSkipped('This code is not used anymore with Symfony 3.1+');
```
GromNaN: This test case must be replaced by a `ArgumentResolverTest` for Symfony 3.1+

src/Silex/ArgumentResolver.php
romainneutron: To avoid constructor signature *à rallonge*

tests/Silex/Tests/ControllerResolverTest.php
romainneutron: fixed

romainneutron:
PR updated, comments addressed

src/Silex/ArgumentResolver.php
stof: We are talking about adding a *second* argument here. It would be better to have the constructor instantiating a usable object

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_script:
     - sh -c "if [ '$SYMFONY_DEPS_VERSION' = '' ]; then sed -i 's/~2\.8|^3\.0/2.8.*@dev/g' composer.json; composer update; fi"
     - composer install
 
-script: phpunit
+script: SYMFONY_DEPRECATIONS_HELPER=weak phpunit
```
stof: why weak ?

stof: btw, please use the ``env: global: ...`` config to set the variable instead. This way, it can be overwritten in the matrix

src/Silex/ArgumentResolver.php
romainneutron: woops, I agree ! I started the implemenation by *extending* Symfony's ArgumentResolver and signature is quite big. I finally went to decoration because the class is final.

sorry :) Gonna update this

src/Silex/ArgumentResolver.php
```diff
@@ -0,0 +1,62 @@
...
+        } elseif (is_object($controller) && !$controller instanceof \Closure) {
+            $r = new \ReflectionObject($controller);
+            $r = $r->getMethod('__invoke');
+        } else {
+            $r = new \ReflectionFunction($controller);
+        }
+
+        foreach ($r->getParameters() as $param) {
+            if ($param->getClass() && $param->getClass()->isInstance($this->app)) {
+                $request->attributes->set($param->getName(), $this->app);
```
stof: -1 for that. Injecting the app in the request attribute was a hacky way (the same than ParamConverters in SensioFrameworkExtraBundle). You should use the Symfony 3.1 extension points instead, by using a ArgumentValueResolver for the application

.travis.yml
romainneutron: weak because if a deprecation is added upstream, Silex build would fail without change in the codebase.

src/Silex/ArgumentResolver.php
stof: btw, doing this will force you to provide a way to register value resolvers, which will then make it available for others :smile:

.travis.yml
stof: and this would mean that we would notice it and fix the deprecations

romainneutron:
PR updated, comments addressed

src/Silex/AppArgumentValueResolver.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $this->app = $app;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(Request $request, ArgumentMetadata $argument)
+    {
+        return $argument->getType() === Application::class;
```
GromNaN: That will return `false` if the type hint is made on a subclass of `Application`.

```php
use Silex\Application;
class MyApp extends Application {}
$app = new MyApp;
$app->get('/', function (MyApp $app) { return 'Hello'; });
$app>->run();
```

This might not be considered as a supported feature, but it is certainly a breaking change.

romainneutron: definitely!

src/Silex/Provider/HttpKernelServiceProvider.php
```diff
@@ -27,8 +35,29 @@ public function register(Container $app)
             return new ControllerResolver($app, $app['logger']);
         };
 
+        $app['argument_metadata_factory'] = function ($app) {
+            return new ArgumentMetadataFactory();
+        };
+        $app['argument_value_resolvers'] = function ($app) {
+            return array(
+                new RequestAttributeValueResolver(),
```
romainneutron: might be improved with https://github.com/symfony/symfony/pull/19011

romainneutron:
PR updated, comments addressed

src/Silex/Provider/HttpKernelServiceProvider.php
skalpa: You services will still be registered when using old Symfony versions, but will just be unusable. I'd recommend that you do something like this instead:
```php
if (Kernel::VERSION_ID >= 30100) {
        $app['argument_metadata_factory'] = function ($app) {
            return new ArgumentMetadataFactory();
        };
        $app['argument_value_resolvers'] = function ($app) {
            return array(...);
        };
        $app['argument_resolver'] = function ($app) {
            return new ArgumentResolver($app['argument_metadata_factory'], $app['argument_value_resolvers']);
        };
}
$app['kernel'] = function ($app) {
    return new HttpKernel($app['dispatcher'], $app['resolver'], $app['request_stack'], isset($app['argument_resolver']) ? $app['argument_resolver'] : null);
};
```

src/Silex/Provider/HttpKernelServiceProvider.php
```diff
@@ -24,11 +33,36 @@ class HttpKernelServiceProvider implements ServiceProviderInterface, EventListen
...
+            $app['argument_metadata_factory'] = function ($app) {
+                return new ArgumentMetadataFactory();
+            };
+            $app['argument_value_resolvers'] = function ($app) {
+                return array(
+                    new RequestAttributeValueResolver(),
+                    new RequestValueResolver(),
+                    new DefaultValueResolver(),
+                    new VariadicValueResolver(),
+                    new AppArgumentValueResolver($app),
```
stof: shouldn't it go before the DefaultValueResolver ?

romainneutron:
PR updated, @stof comment addressed. Add support for Symfony 3.2 btw, since symfony/symfony#19011 is merged

tests/Silex/Tests/ControllerResolverTest.php
stof: no need to skip the test entirely. Just keep it and mark it as ``@group legacy``. This way, the class is tested even when you have 3.1 installed locally

src/Silex/ControllerResolver.php
```diff
@@ -37,6 +37,9 @@ public function __construct(Application $app, LoggerInterface $logger = null)
         parent::__construct($logger);
     }
 
+    /**
+     * This method can be dropped once Symfony 3.0 is not supported anymore.
+     */
```
fabpot: When Symfony 3.0 is not supported anymore, that's the whole class that can be removed, right?

romainneutron: that's right, should I move the comment at class level?

fabpot: yes

romainneutron:
PR updated, comment addressed

fabpot:
Thank you @romainneutron.

MAXakaWIZARD:
@romainneutron Could you please explain, what class should be used instead of `ControllerResolver` since it is deprecated (I use my own resolver which extends this class)?

PR #1388, opened 2016-06-17
Add ControllerServiceProviderInterface for Application::registerAndMount

naderman:
Simplifies the setup for service providers which also provide controllers, which is possible as mentioned in the documentation at [http://silex.sensiolabs.org/doc/master/providers.html](http://silex.sensiolabs.org/doc/master/providers.html). This PR also formalizes this behaviour with an interface that combines Pimple's `ServiceProviderInterface` and Silex's `ControllerProviderInterface`.

If you use quite a few of these in a larger application, your app.php ends up having a lot of

``` php
$app->register($fooProvider = new Bar\FooProvider());
$app->mount('/foo', $fooProvider);
```

instead of which you can use

``` php
$app->registerAndMount('/foo', new Bar\FooProvider());
```

Not sure if you think this is superfluous, but I find this a bit more convient, and it's a rather small addition.

### DIFF
--- a/src/Silex/Api/ControllerServiceProviderInterface.php
+++ b/src/Silex/Api/ControllerServiceProviderInterface.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of the Silex framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Silex\Api;
+
+use Pimple\ServiceProviderInterface;
+
+/**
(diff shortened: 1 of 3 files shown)


ragboyjr:
@naderman, you could just make your service providers apart of the bootable interface and then mount your controllers then.

then you'd only have to use the register method, and when the framework boots, it'll call your provider where you can mount your controllers.

GromNaN:
The solution proposed by @ragboyjr is the one that is implemented by the [WebProfilerServiceProvider](https://github.com/silexphp/Silex-WebProfiler/blob/master/WebProfilerServiceProvider.php#L339)

fabpot:
Closing as apparently that's not something people want.

PR #1397, opened 2016-07-09
Bootstrap composer autoload in PHPUnit

MatthieuMota:
Avoid to require autoload in test class.


fabpot:
Thank you @MatthieuMota.

PR #1398, opened 2016-07-13
Update nginx sample with PHP7 as default

SpacePossum:
welcome PHP7 :)

### DIFF
--- a/doc/web_servers.rst
+++ b/doc/web_servers.rst
@@ -57,9 +57,7 @@ The **minimum configuration** to get your application running under Nginx is:
         # location ~ ^/(index|index_dev)\.php(/|$) {
         location ~ ^/index\.php(/|$) {
             # the ubuntu default
-            fastcgi_pass   unix:/var/run/php/php7.0-fpm.sock;
-            # for running php5
-            # fastcgi_pass   unix:/var/run/php5-fpm.sock;
+            fastcgi_pass   unix:/var/run/php/phpX.X-fpm.sock;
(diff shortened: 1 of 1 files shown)


doc/web_servers.rst
```diff
@@ -57,7 +57,9 @@ The **minimum configuration** to get your application running under Nginx is:
         # location ~ ^/(index|index_dev)\.php(/|$) {
         location ~ ^/index\.php(/|$) {
             # the ubuntu default
-            fastcgi_pass   unix:/var/run/php5-fpm.sock;
+            fastcgi_pass   unix:/var/run/php/php7.0-fpm.sock;
+            # for running php5
+            # fastcgi_pass   unix:/var/run/php5-fpm.sock;
```
fabpot: Instead, I would just replace 5 by X. This is just a template afterall.

SpacePossum: between 5 and 7 the default directory to run the `sock` on changed for Ubuntu from:
`/var/run/` to `var/run/php`
minor diff but still very useful to know imho

fabpot: But this doc does not assume Ubuntu anywhere. Again, this is just a template, not a ready to go configuration.

SpacePossum: Far enough, I was thrown off guard with the current `# the ubuntu default` and the `centos` example :+1:

fabpot:
Thank you @SpacePossum.

PR #1399, opened 2016-07-15
Added argument description to twig asset() method

DaveC49:
I have added a description of the arguments for the asset() method. I didn't find them obvious when I first used the AssetServiceProvider, although it makes sense in hindsight. The description would have saved me some time getting the paths right


fabpot:
Thank you @DaveC49.

PR #1408, opened 2016-08-04
fix lazy evaluation of 'monolog.use_error_handler'

frankhildebrandt:
the value of 'debug' could be set after registering the serviceprovider.


fabpot:
Thank you @frankhildebrandt.

PR #1410, opened 2016-08-05
Update testing.rst - preventing Fatal Error: Class 'Silex\WebTestCase' not found

Czarnodziej:
Taken from:
https://github.com/silexphp/Silex-Skeleton/blob/master/phpunit.xml.dist

Not adding bootstrap segment to phpunit.xml.dist causes Fatal Error: Class 'Silex\WebTestCase' not found


fabpot:
Thank you @Czarnodziej.

PR #1412, opened 2016-08-15
fix header format check strpos argument order

gpedic:
The argument order for strpos in the Guard Authentication example was mixed up


jdreesen:
Duplicate of #1407

PR #1413, opened 2016-08-17
Update Testing.rst - remove duplicate tag in phpunit xml

ncareau:
Removed a duplicate tag (bootstrap) in the phpunit.xml example.


fabpot:
Thank you, I've reverted the culprit PR.

PR #1428, opened 2016-09-26
Add Symfony Twig Brigde installation in asset doc

benjaminjonard:
Add Symfony Twig Brigde installation in asset doc, I have the following error otherwise : `Unknown "asset" function`
Made it clearer that twig-bridge is required.


fabpot:
Thank you @benjaminjonard.

PR #1430, opened 2016-10-03
Test/Explain monolog.exception.logger_filter

SpacePossum:
This PR:
- adds a test for the `monolog.exception.logger_filter` configuration option for the MonologServiceProvider
- updates the description stating it is not a filter, but a mapping function from exception to error level


fabpot:
Thank you @SpacePossum.

PR #1432, opened 2016-10-04
Update session.rst to explain handlers

Shruubi:
Expand the documentation to explain how to use different session providers.


fabpot:
Thank you @Shruubi.

PR #1433, opened 2016-10-06
RequestStack link in Services documentation is broken

carlosbuenosvinos:
There was an extra space breaking the URL formatting


fabpot:
Good catch, thanks @carlosbuenosvinos.

PR #1434, opened 2016-10-09
Added a default kernel.root_dir value

jeremyFreeAgent:
This add a default kernel.root_dir value that can be used by the FileLinkFormater. See https://github.com/silexphp/Silex-WebProfiler/pull/99 for details.

With a default value, the https://github.com/silexphp/Silex-WebProfiler/pull/99 will work out of the box.

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -65,6 +65,7 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
         $this['debug'] = false;
         $this['charset'] = 'UTF-8';
         $this['logger'] = null;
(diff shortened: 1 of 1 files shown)


jeremyFreeAgent:
I don't really like the `dirname` calls but I don't see other way to configure a default value. If someone has a idea, I'll take it!

SpacePossum:
I don't see why this is needed as it is going to call 6 times `dirname` each time the Application is created, than assumes there is an `app` folder which is not standard for Silex, just to guess a default for a debug-profiler which shouldn't be not loaded during production.

I think it would be better to add it to the doc's for the profiler how to do the setup

```php
$this['kernel.root_dir'] = function() {
    return _DIR__.'/../../../../app';
}
```

jeremyFreeAgent:
Closed since the logic of the baseDir changed.

PR #1438, opened 2016-10-26
Monolog doc: Use PSR-3 methods

Tobion:
The other methods are not PSR-3 standard and will be removed in monolog 2.


fabpot:
Thank you @Tobion.

PR #1441, opened 2016-10-30
fix typo in SecurityServiceProvider exception message

angyvolin:


fabpot:
Thank you @angyvolin.

PR #1444, opened 2016-11-06
Use DebugProcessor instead of DebugHandler when available in MonologServiceProvider

nicolas-grekas:
Related to https://github.com/symfony/symfony/pull/20416


src/Silex/Provider/MonologServiceProvider.php
```diff
@@ -62,8 +63,12 @@ public function register(Container $app)
 
             $log->pushHandler($handler);
 
-            if ($app['debug'] && isset($app['monolog.handler.debug'])) {
-                $log->pushHandler($app['monolog.handler.debug']);
+            if ($app['debug'] && $bridge) {
```
fabpot: Are you sure it works? `$bridge` is not defined in this scope.

nicolas-grekas: oups, indeed, fixed

fabpot:
Thank you @nicolas-grekas.

PR #1449, opened 2016-11-15
The services must be set only if HttpKernelRuntime exists

olivier34000:
Some services are broken if HttpKernelRuntime does not exist.

### DIFF
--- a/tests/Silex/Tests/Provider/TwigServiceProviderTest.php
+++ b/tests/Silex/Tests/Provider/TwigServiceProviderTest.php
@@ -116,4 +116,13 @@ class TwigServiceProviderTest extends \PHPUnit_Framework_TestCase
 
         $this->assertInstanceOf('Twig_Environment', $app['twig']);
     }
+
+    public function testServices()
+    {
+        $app = new Application();
+        $app->register(new TwigServiceProvider());
+        foreach ($app->keys() as $id) {
+            $app[$id];
+        }
+    }
 }
--- a/src/Silex/Provider/TwigServiceProvider.php
+++ b/src/Silex/Provider/TwigServiceProvider.php
@@ -122,6 +122,21 @@ class TwigServiceProvider implements ServiceProviderInterface
                 }
 
                 if (class_exists(HttpKernelRuntime::class)) {
+                    $app['twig.runtime.httpkernel'] = function ($app) {
+                        return new HttpKernelRuntime($app['fragment.handler']);
+                    };
+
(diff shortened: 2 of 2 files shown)


tests/Silex/Tests/Provider/TwigServiceProviderTest.php
```diff
@@ -116,4 +116,13 @@ public function testFormWithoutCsrf()
 
         $this->assertInstanceOf('Twig_Environment', $app['twig']);
     }
+
+    public function testServices()
+    {
+        $app = new Application();
+        $app->register(new TwigServiceProvider());
+        foreach ($app->keys() as $id) {
+            $app[$id];
```
fabpot: This test does not test anything useful as services are instantiated only on-demand.

fabpot:
Can you explain when the current code breaks?

tests/Silex/Tests/Provider/TwigServiceProviderTest.php
olivier34000: The test is here only to check that the different services can be created without errors. If you execute the test on the previous version the test is broken. I think it's not a good practice to be able to create services even if the services are instantiated on-demand. 
In my codebase we check all the services can be created and since the last stable version is does not work anymore.

skalpa:
The code doesn't break as you correctly surrounded the call to `addRuntimeLoader` with an `if (class_exists(HttpKernelRuntime::class))`.

The problem is that the actual code registers services that may be unusable, and that should be avoided: there are valid use cases for wanting to be able to blindly instantiate all the registered services (i.e: a tool that dumps the content of the container to generate an `array('service' => 'Class')` lookup table for IDEs).

So, I'm 👍 regarding the change to `TwigServiceProvider`, but also thinks that the actual test shouldn't be added. If a test has to be added, it should just assert that the services that depend on `HttpKernelRuntime` only get registered when the class is available, using something like:

```php
   $this->assertSame(isset($app['twig.runtime.httpkernel']), class_exists(HttpKernelRuntime::class));
```

olivier34000:
I close the PR because I have no more the issue with the last versions

IMM0rtalis:
But this is still broken. 
I use PimpleDumpProvider which tries to instantiate all containers to create a file for phpstorm. 
So I got:
`Attempted to load class "HttpKernelRuntime" from namespace "Symfony\Bridge\Twig\Extension".
Did you forget a "use" statement for another namespace?`
HttpKernelRuntime looks like it belongs to twig-bridge so it should be inside the condition a few lines above.

PR #1452, opened 2016-11-21
[WIP] Update form.rst

SpacePossum:
* no need to add `symfony/translation` as dep. as the `symfony/validator` has it as requirement
* you can use form with Twig without `symfony/config` and `symfony/translation`

### DIFF
--- a/doc/providers/form.rst
+++ b/doc/providers/form.rst
@@ -132,14 +132,20 @@ form by adding constraints on the fields::
         'translator.domains' => array(),
     ));
 
-    $form = $app['form.factory']->createBuilder('form')
-        ->add('name', 'text', array(
+    // some default data for when the form is displayed the first time
+    $data = array(
+        'name' => 'Your name',
+        'email' => 'Your email',
+    );
+        
+    $form = $app['form.factory']->createBuilder(FormType::class, $data)
+        ->add('name', TextType::class, array(
             'constraints' => array(new Assert\NotBlank(), new Assert\Length(array('min' => 5)))
         ))
-        ->add('email', 'text', array(
+        ->add('email', EmailType::class, array(
             'constraints' => new Assert\Email()
         ))
-        ->add('billing_plan', 'choice', array(
+        ->add('billing_plan', ChoiceType::class, array(
             'choices' => array(1 => 'free', 2 => 'small_business', 3 => 'corporate'),
             'expanded' => true,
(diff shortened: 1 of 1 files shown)


doc/providers/form.rst
```diff
@@ -53,12 +53,12 @@ Registering
         composer require symfony/form
 
     If you are going to use the validation extension with forms, you must also
-    add a dependency to the ``symfony/config`` and ``symfony/translation``
```
fabpot: Are you sure that translation is not a requirement? Or is it a transitive requirement of validator?

SpacePossum: it transitive since [v2.2.0](https://packagist.org/packages/symfony/validator#v2.2.0) of the Validator
 (I found out why I didn't need the translation provider, I've a custom `trans`-filter I my project to bypass the need for the `symfony/translation` package)

SpacePossum:
hmm lot of unrelated tests are failing, is this known?

doc/providers/form.rst
```diff
@@ -132,15 +132,21 @@ form by adding constraints on the fields::
         'translator.domains' => array(),
     ));
 
-    $form = $app['form.factory']->createBuilder('form')
```
fabpot: The updated example does not work for older versions of Symfony. I think we need to keep both examples and tell the user which version to use depending on the version of Symfony they have.

fabpot:
I've fixed the deprecation notices, so tests should be green again when you update this PR.

doc/providers/form.rst
```diff
@@ -132,15 +132,21 @@ form by adding constraints on the fields::
...
             'constraints' => array(new Assert\NotBlank(), new Assert\Length(array('min' => 5)))
         ))
-        ->add('email', 'text', array(
+        ->add('email', EmailType::class, array(
             'constraints' => new Assert\Email()
         ))
-        ->add('billing_plan', 'choice', array(
-            'choices' => array(1 => 'free', 2 => 'small_business', 3 => 'corporate'),
+        ->add('billing_plan', ChoiceType::class, array(
+            'choices' => array('free' => 1, 'small business' => 2, 'corporate' => 3),
```
fabpot: All these changes are only for a recent version of Symfony. So, we need to keep both and indicate when to use which.

SpacePossum:
I'll re-open when I did the rewrite, thanks!

PR #1453, opened 2016-11-21
twig.rst Add missing trait method to doc

SpacePossum:


fabpot:
Thank you @SpacePossum.

GromNaN:
Is it necessary to maintain such documentation while we could link/include the API doc? http://silex.sensiolabs.org/api/master/Silex/Application/TwigTrait.html

fabpot:
Not sure people are using the API docs. Personally, I don't, I prefer to read a "real" doc with "real" paragraph and sentences :)

PR #1455, opened 2016-11-23
Syntax error in documentaiton Security Guard

pierreboissinot:
"=>" replaced by "="


fabpot:
Good catch, thanks @pierreboissinot.

PR #1456, opened 2016-11-29
Make it easier to register SwiftMailer plugins

c960657:
This patch makes it easier to register SwiftMailer plugins.

### DIFF
--- a/src/Silex/Provider/SwiftmailerServiceProvider.php
+++ b/src/Silex/Provider/SwiftmailerServiceProvider.php
@@ -71,17 +71,6 @@ class SwiftmailerServiceProvider implements ServiceProviderInterface, EventListe
             $transport->setPassword($options['password']);
             $transport->setAuthMode($options['auth_mode']);
 
-            if (null !== $app['swiftmailer.sender_address']) {
-                $transport->registerPlugin(new \Swift_Plugins_ImpersonatePlugin($app['swiftmailer.sender_address']));
-            }
-
-            if (!empty($app['swiftmailer.delivery_addresses'])) {
-                $transport->registerPlugin(new \Swift_Plugins_RedirectingPlugin(
-                    $app['swiftmailer.delivery_addresses'],
-                    $app['swiftmailer.delivery_whitelist']
-                ));
-            }
-
             return $transport;
         };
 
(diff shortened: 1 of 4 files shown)


tests/Silex/Tests/Provider/SpoolStub.php
```diff
@@ -36,7 +36,7 @@ public function isStarted()
 
     public function queueMessage(\Swift_Mime_Message $message)
     {
-        $this->messages[] = $message;
+        $this->messages[] = clone $message;
```
fabpot: Should probably be part of another PR as it does not seems related and not tested.

src/Silex/Provider/SwiftmailerServiceProvider.php
```diff
@@ -97,13 +86,34 @@ public function register(Container $app)
...
+            }
+
+            foreach ($plugins as $plugin) {
+                $dispatcher->bindEventListener($plugin);
+            }
+
+            return $dispatcher;
         };
 
+        $app['swiftmailer.plugins'] = array();
```
fabpot: This should not be an array as per your example in the docs, the plugins would not be lazy-loaded. Instead, you should do the same as for `validator.object_initializers` for instance:

```php
$app['swiftmailer.plugins'] = function ($app) {
    return array();
}
```

c960657: Fixed.

tests/Silex/Tests/Provider/SpoolStub.php
c960657: Without this, the new `testSwiftMailerSenderAddress` fails.

[Swift_Plugins_ImpersonatePlugin](https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Plugins/ImpersonatePlugin.php) reverts the headers after the message has been sent (in the `sendPerformed` event), so without cloning, `$message->getReturnPath()` would not return the value used when the message is sent.

src/Silex/Provider/SwiftmailerServiceProvider.php
fabpot: Does not look like it is fixed :)

tests/Silex/Tests/Provider/SpoolStub.php
fabpot: ok

src/Silex/Provider/SwiftmailerServiceProvider.php
c960657: Arg! Fixed "again" :-)

fabpot:
Thank you @c960657.

PR #1457, opened 2016-11-29
Fix VarDumping

SpacePossum:
To reproduce

composer.json
```json
{
  "require": {
    "silex/silex": "v2.0.2"
  },
  "require-dev": {
    "symfony/var-dumper": "v3.1.2"
  }
}

```

index.php
```php
<?php
require_once __DIR__.'/../vendor/autoload.php';

$app = new Silex\Application();
$app['debug'] = true;
$app->register(new Silex\Provider\VarDumperServiceProvider());

$app->get('/', function() {
    dump(false);
    return '1';
});

$app->run();
```

Output (in browser) does not contain the `dump` output, as it is outputted for CLI.
This only happens when `$app['debug']` is set to `true`.
This PR addresses that in a way the service provider becomes more independent of the `Application` execution, which is in line with most providers being aware of a `Container` but being framework agnostic.

Other issue addressed is to allow for different output selection for the `dump` function.
Kindly note that this doesn't work yet when using `dump` in a template when using the `TwigServiceProvider` because [DumpExtension](https://github.com/symfony/symfony/blob/v3.1.7/src/Symfony/Bridge/Twig/Extension/DumpExtension.php#L69) changes the output for some reason.

(last thingy is a dead import removed)

cc @tiemevanveen as he ran into this issue before me

### DIFF
--- a/src/Silex/Provider/VarDumperServiceProvider.php
+++ b/src/Silex/Provider/VarDumperServiceProvider.php
@@ -55,7 +55,7 @@ class VarDumperServiceProvider implements ServiceProviderInterface, BootableProv
         // This code is here to lazy load the dump stack.
         VarDumper::setHandler(function ($var) use ($app) {
             $handler = function ($var) use ($app) {
-                $env = null === $app['var_dumper.env'] ? PHP_SAPI : $app['var_dumper.env'];
+                $env = null === $app['var_dumper.env'] ? \PHP_SAPI : $app['var_dumper.env'];
                 'cli' === $env
                     ? $app['var_dumper.cli_dumper']->dump($app['var_dumper.cloner']->cloneVar($var))
                     : $app['var_dumper.html_dumper']->dump($app['var_dumper.cloner']->cloneVar($var))
--- a/src/Silex/Provider/VarDumperServiceProvider.php
+++ b/src/Silex/Provider/VarDumperServiceProvider.php
@@ -15,6 +15,7 @@ use Pimple\Container;
 use Pimple\ServiceProviderInterface;
 use Silex\Application;
 use Silex\Api\BootableProviderInterface;
+use Symfony\Component\VarDumper\Dumper\HtmlDumper;
 use Symfony\Component\VarDumper\VarDumper;
 use Symfony\Component\VarDumper\Cloner\VarCloner;
 use Symfony\Component\VarDumper\Dumper\CliDumper;
@@ -32,6 +33,10 @@ class VarDumperServiceProvider implements ServiceProviderInterface, BootableProv
             return new CliDumper($app['var_dumper.dump_destination'], $app['charset']);
         };
 
+        $app['var_dumper.html_dumper'] = function ($app) {
+            return new HtmlDumper($app['var_dumper.dump_destination'], $app['charset']);
+        };
+
         $app['var_dumper.cloner'] = function ($app) {
             return new VarCloner();
         };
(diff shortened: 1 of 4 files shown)


src/Silex/Provider/VarDumperServiceProvider.php
```diff
@@ -45,13 +50,14 @@ public function boot(Application $app)
...
-        // This code is here to lazy load the dump stack. This default
-        // configuration for CLI mode is overridden in HTTP mode on
-        // 'kernel.request' event
+        // This code is here to lazy load the dump stack.
         VarDumper::setHandler(function ($var) use ($app) {
-            VarDumper::setHandler($handler = function ($var) use ($app) {
-                $app['var_dumper.cli_dumper']->dump($app['var_dumper.cloner']->cloneVar($var));
-            });
+            $handler = function ($var) use ($app) {
+                'cli' === PHP_SAPI
```
SpacePossum: while optimal, this makes it also near impossible to utest as the value of PHP_SAPI cannot be changed @ runtime.

tiemevanveen:
Thanks for reporting my problem! :+1: +1

SpacePossum:
ping @fabpot anything you want changed here?

src/Silex/Provider/VarDumperServiceProvider.php
fabpot: Isn't possible to define a `Silex\Provider::PHP_SAPI` constant in the test? I think it should work.

src/Silex/Provider/VarDumperServiceProvider.php
```diff
@@ -15,6 +15,7 @@
 use Pimple\ServiceProviderInterface;
 use Silex\Application;
 use Silex\Api\BootableProviderInterface;
+use Symfony\Component\VarDumper\Dumper\HtmlDumper;
```
nicolas-grekas: I'd move this clear CliDumper

nicolas-grekas:
Just to be sure to understand: right now, the dump output is always sent to STDERR, and you'd like it to go in HTTP content instead?
In Symfony, we have the same `PHP_SAPI` based default, but very soon, once a request comes in, we force HtmlDumper - and if a Command starts, we force CliDumper.

src/Silex/Provider/VarDumperServiceProvider.php
GromNaN: This could be configured with a default value in `register` method.

```php
function register(Container $app) {
    ...
    $app['var_dumper.is_cli'] = PHP_SAPI;
    ...
}
```

SpacePossum:
@nicolas-grekas correct, I would like the Vardumper to dump to HTTP output and format in HTML format when used in a controller (for example) and not only when used in Twig.

Silex doesn't has/supports Commands OOTB and most of the providers are framework agnostic (i.e. depend on the container only so the provider can be used without the Silex/Application and/or its middleware), I hope this PR makes the provider a better fit for that.

SpacePossum:
@nicolas-grekas can I bother you with a request to review again?

fabpot:
@nicolas-grekas Can you have a look at this PR? Thanks.

SpacePossum:
ping @nicolas-grekas :)

SpacePossum:
Failing test is not related, it has to do with changes in PHPUnit (5 vs. 6) in Travis env.

SpacePossum:
I'm closing my PR's I lost interest in as I do not want to leave the expression that I'll follow up on those.
Thanks to all the reviewers!

PR #1466, opened 2016-12-21
Change 'swiftmailer.plugins' to Pimple container

c960657:
In #1456, it was made possible to register Swiftmailer plugins by populating `$app['swiftmailer.plugins']`.

In order to make it easy for other providers to add plugins, I suggest changing `$app['swiftmailer.plugins']` to a Pimple container. This allows a looser coupling than $app->extend() that requires the providers to be registered in a specific correct order.

### DIFF
--- a/src/Silex/Provider/SwiftmailerServiceProvider.php
+++ b/src/Silex/Provider/SwiftmailerServiceProvider.php
@@ -91,26 +91,28 @@ class SwiftmailerServiceProvider implements ServiceProviderInterface, EventListe
 
             $plugins = $app['swiftmailer.plugins'];
 
-            if (null !== $app['swiftmailer.sender_address']) {
-                $plugins[] = new \Swift_Plugins_ImpersonatePlugin($app['swiftmailer.sender_address']);
-            }
-
-            if (!empty($app['swiftmailer.delivery_addresses'])) {
-                $plugins[] = new \Swift_Plugins_RedirectingPlugin(
-                    $app['swiftmailer.delivery_addresses'],
-                    $app['swiftmailer.delivery_whitelist']
-                );
-            }
-
-            foreach ($plugins as $plugin) {
-                $dispatcher->bindEventListener($plugin);
+            foreach ($plugins->keys() as $key) {
+                $dispatcher->bindEventListener($plugins[$key]);
             }
 
             return $dispatcher;
         };
 
         $app['swiftmailer.plugins'] = function ($app) {
-            return array();
+            $plugins = new Container();
+
+            if (null !== $app['swiftmailer.sender_address']) {
+                $plugins['sender_address'] = new \Swift_Plugins_ImpersonatePlugin($app['swiftmailer.sender_address']);
+            }
+
+            if (!empty($app['swiftmailer.delivery_addresses'])) {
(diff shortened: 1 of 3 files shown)


src/Silex/Provider/SwiftmailerServiceProvider.php
```diff
@@ -91,26 +91,30 @@ public function register(Container $app)
 
             $plugins = $app['swiftmailer.plugins'];
 
+            $keys = $plugins->keys();
+            sort($keys);
```
stof: why sorting by key ?

hkdobrev: @c960657 I am using Silex 2.0.x and I've added an array of swiftmailer plugins to Pimple and I'm registering them lazily. The order I've added them is important for me, though. I have a plugin which for specific emails would add a bcc and another plugin which would blacklist/whitelist certain emails and domains based on environment. So if I register the filtering plugin first and it filters out all emails, the plugin adding bcc would not work, because the filtering plugin will cancel the email event. If I add the bcc plugin first, and it adds a whitelisted email, the filtering plugin will remove all emails and just send an email to the bcc email without canceling the whole email. I don't see how sorting implicitly works towards more predictability.

Other than that, thank you for work on improving the plugins configuration, it is much appreciated!

stof: Keeping the registration order (which is what ``$plugins->keys()`` does) looks better indeed

c960657: > I don't see how sorting implicitly works towards more predictability.

I imagined something like this, similar to [conf.d](http://superuser.com/questions/705297/in-what-order-does-apache-load-conf-files-and-which-ones/705317#705317):
```
$app['swiftmailer.plugins']['001-bcc] = ... plugin that should run first ...
$app['swiftmailer.plugins']['002-filter'] = ... plugin that should run later ...
$app['swiftmailer.plugins']['foo'] = ... plugin where order does not matter  ...
```
The use-case for this is if multiple providers register plugins where the plugin order is important. If you register multiple plugins in your own code within a single function, I agree that keeping the registration order is more intuitive.

Anyway, I don't have strong feelings about this, so I will remove it.

c960657: Now it's removed.

src/Silex/Provider/SwiftmailerServiceProvider.php
```diff
@@ -91,26 +91,28 @@ public function register(Container $app)
...
             $plugins = $app['swiftmailer.plugins'];
 
+            foreach ($plugins->keys() as $key) {
+                $dispatcher->bindEventListener($plugins[$key]);
+            }
+
+            return $dispatcher;
+        };
+
+        $app['swiftmailer.plugins'] = function ($app) {
```
GromNaN: Is it necessary to lazy-load the swiftmailer.plugins container ?
If we follow your example, it would be initialized before the boot of the application.

c960657:
On second thought, this suggestion seems to go against the Silex convention of defining such arrays as `$app['foo'] = function () { return array(); }` and extend them using `$app->extend('foo', function () { … })`.

I still thing the Container way has its merits, but let's stay consistent. I'll close this ticket.

PR #1467, opened 2016-12-28
Include LocaleServiceProvider in Form doc

EdwinKortman:
The translation service requires locale service to be loaded first. The form documentation is unclear on this point. :)


fabpot:
It's not a requirement. As the exception message says: you must either load this service provider or define the locale parameter.

PR #1473, opened 2017-01-08
Added the FormRegistry as a service to enable the extension point

HeahDude:
Closes https://github.com/symfony/symfony/issues/21199.

This PR makes the form configuration aligned with the symfony full stack, see https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Resources/config/form.xml#L11-L29.


fabpot:
Thank you @HeahDude.

PR #1478, opened 2017-01-21
Update web_servers.rst

adamcameron:
Just another option for Apache that works


fabpot:
Thank you @adamcameron.

PR #1484, opened 2017-02-08
Added option validator.mapping.use_annotation

jlHertel:
Added option validator.mapping.use_annotation to change the default usage of StaticMethodLoader to AnnotationLoader.

Added option validator.mapping.cache to allow passing a CacheInterface for usage in the mapping loader.

This should simplify the creation of a validator that reads annotations instead of static methods.

### DIFF
--- a/src/Silex/Provider/ValidatorServiceProvider.php
+++ b/src/Silex/Provider/ValidatorServiceProvider.php
@@ -18,6 +18,8 @@ use Symfony\Component\Validator\Validator;
 use Symfony\Component\Validator\Mapping\Factory\LazyLoadingMetadataFactory;
 use Symfony\Component\Validator\Mapping\Loader\StaticMethodLoader;
 use Symfony\Component\Validator\Validation;
+use Doctrine\Common\Annotations\AnnotationReader;
+use Symfony\Component\Validator\Mapping\Loader\AnnotationLoader;
 
 /**
  * Symfony Validator component Provider.
@@ -46,7 +48,21 @@ class ValidatorServiceProvider implements ServiceProviderInterface
         };
 
         $app['validator.mapping.class_metadata_factory'] = function ($app) {
-            return new LazyLoadingMetadataFactory(new StaticMethodLoader());
+            $loader = null;
+
+            if ($app->offsetExists('validator.mapping.use_annotation') && $app['validator.mapping.use_annotation'] === true) {
+                $loader = new AnnotationLoader(new AnnotationReader());
+            } else {
+                $loader = new StaticMethodLoader();
+            }
+
+            $cache = null;
+
+            if ($app->offsetExists('validator.mapping.cache')) {
+                $cache = $app['validator.mapping.cache'];
+            }
+
+            return new LazyLoadingMetadataFactory($loader, $cache);
         };
 
         $app['validator.validator_factory'] = function () use ($app) {
--- a/tests/Silex/Tests/Provider/ValidatorServiceProviderTest.php
+++ b/tests/Silex/Tests/Provider/ValidatorServiceProviderTest.php
@@ -11,6 +11,7 @@
 
 namespace Silex\Tests\Provider;
 
+use Doctrine\Common\Cache\ArrayCache;
 use Silex\Application;
 use Silex\Provider\TranslationServiceProvider;
 use Silex\Provider\ValidatorServiceProvider;
@@ -55,6 +56,21 @@ class ValidatorServiceProviderTest extends \PHPUnit_Framework_TestCase
         return $app;
     }
 
+    public function testRegisterWithAnnotationLoader()
+    {
+        $app = new Application();
+
+        $app['validator.mapping.use_annotation'] = true;
+
+        $app['validator.mapping.cache'] = function () {
+            return new ArrayCache();
+        };
+
+        $app->register(new ValidatorServiceProvider());
+
+        return $app;
+    }
+
     /**
      * @depends testRegisterWithCustomValidators
      */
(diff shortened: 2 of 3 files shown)


src/Silex/Provider/ValidatorServiceProvider.php
```diff
@@ -46,7 +48,21 @@ public function register(Container $app)
         };
 
         $app['validator.mapping.class_metadata_factory'] = function ($app) {
-            return new LazyLoadingMetadataFactory(new StaticMethodLoader());
+            $loader = null;
```
SpacePossum: No need to set to `null`, it will be set below anyway.
Maybe tests should be added and doc's been updated?
Would like to see this in :)

jlHertel:
@SpacePossum, Ok, I will write some tests and a documentation update. I'm not very sure how to use github to do that in this same pull request. Any idea? Or should I make a new pull request?

jdreesen:
> I'm not very sure how to use github to do that in this same pull request. Any idea? Or should I make a new pull request?

Just push to your branch and GitHub will update the according pull request automatically ;)

src/Silex/Provider/ValidatorServiceProvider.php
jlHertel: Ok, I've added a test case for the setup and updated the docs.
The builds seems to be failing, but as I have seen, this is caused for using a PHPUnit 6+ using the the old `\PHPUnit_Framework_TestCase` syntax.

jlHertel:
Any thoughts on this?

src/Silex/Provider/ValidatorServiceProvider.php
mvanbaak: If instead of `null` you set it to `new StaticMethodLoader` here you can remove the whole `else` class later.

SpacePossum: the `else` performance better than always creating an instance of the `StaticMethodLoader` and than not use it if annotation usage is configured

src/Silex/Provider/ValidatorServiceProvider.php
```diff
@@ -46,7 +48,21 @@ public function register(Container $app)
...
+
+            if ($app->offsetExists('validator.mapping.use_annotation') && $app['validator.mapping.use_annotation'] === true) {
+                $loader = new AnnotationLoader(new AnnotationReader());
+            } else {
+                $loader = new StaticMethodLoader();
+            }
+
+            $cache = null;
+
+            if ($app->offsetExists('validator.mapping.cache')) {
```
skalpa: 1. You shouldn't use `$app->offsetExists('id')` to test if an entry exists 
but `isset($app['id'])`
2. You should always assign a default value to the parameters/services you 
are going to use. This way, you won't have to check if they have been 
defined whenever you need them:

```php
$app['validator.mapping.cache'] = null;

$app['validator.mapping.class_metadata_factory'] = function ($app) {
    // ...

    return new LazyLoadingMetadataFactory($loader, $app['validator.mapping.cache']);
};
```

skalpa:
Why just add a flag and not introduce a new `validator.mapping.loader` 
service that users could override?

- This would let developers use any of the available loaders
- This would also let them configure the `AnnotationReader` they
want to use

Something like this would be better IMHO:

```php
$app['validator.mapping.cache'] = null;

$app['validator.mapping.loader'] = function ($app) {
    return new StaticMethodLoader();
}

$app['validator.mapping.class_metadata_factory'] = function ($app) {
    return new LazyLoadingMetadataFactory($app['validator.mapping.loader'], $app['validator.mapping.cache']);
};
```

If you wish, you could then add an example in the documentation to show how 
to use the `AnnotationLoader` by overriding the new service:

```php
$app->register(new ValidatorServiceProvider(), array(
    'validator.mapping.loader' => function ($app) {
        return new AnnotationLoader($app['annotations']);
    },
));
```

tests/Silex/Tests/Provider/ValidatorServiceProviderTest.php
```diff
@@ -55,6 +56,21 @@ public function testRegisterWithCustomValidators()
         return $app;
     }
 
+    public function testRegisterWithAnnotationLoader()
```
skalpa: Your test doesn't actually test anything.

src/Silex/Provider/ValidatorServiceProvider.php
jlHertel: 1. I think code preference doesn't add anything. But if you read carefully, you will notice that the method `offsetExists` doesn't try to retrieve the value, just check if its there.

2. Well, yes, A default value won't hurt, but in this case I want it to be null anyway, so it doesn't make much sense to me.

tests/Silex/Tests/Provider/ValidatorServiceProviderTest.php
jlHertel: They test if there is some kind of error while registering the service.
Just like many other tests do around in the framework.

jlHertel: About adding this as a new option instead of a service: Its possible to override it already, the documentation doesn't state it.
I have added an option only for commodity and because I think more users would benefit from an option that is so common, instead of overwriting the whole service.

hkdobrev:
@jlHertel Do you think you'd be able to update this soon and resolve the conflicts? Thanks!

hkdobrev:
I guess this should be closed as well as per https://github.com/silexphp/Silex/issues/1588#issuecomment-368390861

fabpot:
@hkdobrev That's correct. Let's close.

Issue #1483, opened 2017-02-07
Incompatible with PHPUnit 6.*


sneps85:
Actual version is not compatible with PHPUnit 6.* as \PHPUnit_Framework_TestCase now uses namespaces. Also, PHPUnit 6.* requires PHP 7

ry167:
This actually was a huge pain. I was getting `PHP Fatal error:  Class 'PHPUnit_Framework_TestCase' not found in {PROJECT_DIR}/vendor/silex/silex/src/Silex/WebTestCase.php on line 22` as the error from this.

Change your `composer.json` lien for PHPUnit to `"phpunit/phpunit": "~5.7"`

Silex has `WebTestCase` which depends on PHPUnit, so it probably should be in `require-dev` for Silex...

araneta:
i tried to use phpunit 5.7 with php7 and i got this message when running phpunit . : No tests executed!
tests are not working. what should i do?

ry167:
@araneta Was `No tests executed` the only error message you received?

jeanpralo:
Definitely a big of a headache.

Related problem on symfony:
https://github.com/symfony/symfony/issues/21534

adamcameron:
I think short-term Silex ought to be patched to restrict the PHPUnit version to 5.x (ie: via its composer.json). That said, using 5.x is no use to me... I need to be using 6.x as the rest of my tests are 6.x.

No worries, I can work around it. Just thought it worthwhile adding a "me too" here. Sorry :-S

curry684:
@adamcameron my workaround at https://github.com/symfony/symfony/issues/21534#issuecomment-278278352 is cleaner and more resilient. You can just put it in phpunit's bootstrap file in your own project.

adamcameron:
@curry684 cool: nice one!

fabpot:
Closing this one as PHPUnit won't be added to the `require` section of Silex as it's a soft dependency. The workaround mentioned by @curry684 looks the way to go.

adamcameron:
Yeah entirely fair enough. Cheers!

sinasalek:
There is also another workaround using composer directly. there is a plugin for class aliasing. This way there no need to change anything else. it can also be added to silex composer till it is fixed properly
https://github.com/TYPO3/class-alias-loader

PR #1487, opened 2017-02-14
Updated composer.json to include phpunit/phpunit ~5.7 as a dev requirement

ry167:
`PHPUnit_Framework_TestCase` from PHPUnit is required in `WebTestCase.php` and Silex includes a `phpunit.xml.dist`

PHPUnit 6.0 which is now the main stable release is not compatible with Silex due to the removal of `PHPUnit_Framework_TestCase` in favor of `PHPUnit\Framework\TestCase`

This resolves issue #1483 and partially re-implements pull request #1075.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,8 @@
         "twig/twig": "~1.28|~2.0",
         "doctrine/dbal": "~2.2",
         "swiftmailer/swiftmailer": "~5",
-        "monolog/monolog": "^1.4.1"
+        "monolog/monolog": "^1.4.1",
+        "phpunit/phpunit": "~5.7"
     },
     "replace": {
         "silex/api": "self.version",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -47,7 +47,8 @@
         "twig/twig": "~1.28|~2.0",
         "doctrine/dbal": "~2.2",
         "swiftmailer/swiftmailer": "~5",
-        "monolog/monolog": "^1.4.1"
+        "monolog/monolog": "^1.4.1",
+        "phpunit/phpunit": "~5.7"
```
stof: this must also allow PHPUnit 4.8, as PHPUnit 5 does not support PHP 5.5

ry167: I'm not super familiar with defining requirements for different versions of PHP. 

Should `"phpunit/phpunit": "~4.8|~5.7"` fix this?

stof: yes

ry167: Thanks :+1:

GromNaN:
What about using `simple-phpunit` instead? Like for Pimple: https://github.com/silexphp/Pimple/pull/218

fabpot:
Closing in favor of #1505

PR #1490, opened 2017-02-19
Update README.rst

Narno:
Add code-bloc for PHP and delete `$` before command lines.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,7 @@ The recommended way to install Silex is through `Composer`_:
 
 .. code-block:: bash
 
-    $ composer require silex/silex "~2.0"
+    composer require silex/silex "~2.0"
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -25,7 +27,7 @@ The recommended way to install Silex is through `Composer`_:
 
 .. code-block:: bash
 
-    composer require silex/silex "~2.0"
+    $ composer require silex/silex "~2.0"
```
fabpot: Actually, I would prefer that we remove `$` elsewhere to avoid bad copy/paste.

xabbuh: Isn't it possible to reuse the `terminal` code block that is used for the Symfony documentation?

fabpot:
Thank you @Narno.

Issue #1486, opened 2017-02-13
SwiftmailerServiceProvider configuration


ramadom:
Silex\Provider\SwiftmailerServiceProvider resets configuration parameters that were set before register. $app['swiftmailer.options'] becomes empty after SwiftmailerProvider is registered.

```
class SwiftmailerServiceProvider implements ServiceProviderInterface, EventListenerProviderInterface
{
    public function register(Container $app)
    {
        $app['swiftmailer.options'] = array();  // Works correctly if this line is removed

```

mathroc:
this is the expected behavior for all silex providers. you should override providers configurations *after* registering them (or with the second argument of `$app->register()`)

stof:
and this is the behavior of **all** Silex providers

ramadom:
Yes, all providers reset options..  To use existing options would seem like a nicer behavior, but if it is expected it means it is correct.

PR #1491, opened 2017-03-14
Remove overwriting of $app['swiftmailer.options']

melz:
When registering the SwiftMailerServiceProvider, $app['swiftmailer.options'] is immediately reset to an empty array so any configuration has to be loaded after the provider has been registered (which is not the same behavior as the other providers - at least of the ones I have used so far).

In line 64, we are expecting the same array to be populated and replaced into the final options array so it seems odd that we would want to 'initialize' that array at the beginning of the registration call:

```
 $options = $app['swiftmailer.options'] = array_replace(array(
                'host' => 'localhost',
                'port' => 25,
                'username' => '',
                'password' => '',
                'encryption' => null,
                'auth_mode' => null,
            ), $app['swiftmailer.options']);
```

### DIFF
--- a/src/Silex/Provider/SwiftmailerServiceProvider.php
+++ b/src/Silex/Provider/SwiftmailerServiceProvider.php
@@ -28,7 +28,6 @@ class SwiftmailerServiceProvider implements ServiceProviderInterface, EventListe
 {
     public function register(Container $app)
     {
-        $app['swiftmailer.options'] = array();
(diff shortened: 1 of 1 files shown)


hkdobrev:
Please see #1486 and https://github.com/silexphp/Pimple/pull/212.

fabpot:
Closing for the reasons given in the linked issues/PRs

PR #1494, opened 2017-03-21
Flush the routes on boot

bashilbers:
This allows for better middleware since the routeCollection is exposed early on compared to (currently) only being available after handling a request. Middleware authors that depend on the RouteCollection used to depend on the third party's ability to call the flush() method manually, which is not wanted.

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -105,6 +105,9 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
         }
 
         $this->booted = true;
+        
+        // Flush the routes to a RouteCollection
+        $this->flush();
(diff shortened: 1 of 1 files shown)


bashilbers:
Guys?

GromNaN:
Some providers adds routes to the application in the boot method. If the flush method is called before, these route are missing. The SecurityProvider is an example: https://github.com/silexphp/Silex/blob/master/src/Silex/Provider/SecurityServiceProvider.php#L677

If you rebase your branch, you may see some tests fail.

bashilbers:
You are right! I guess I'll have to find a workaround in my middleware.

PR #1501, opened 2017-04-14
Add PHP 7.1 to the build

emirb:
Tests on >=7.0 are depending on PHPUnit version lock, this requires  https://github.com/silexphp/Silex/pull/1487 merging first.


emirb:
Done in https://github.com/silexphp/Silex/pull/1505

PR #1503, opened 2017-04-16
Added more options to security.firewalls

K0nias:
Added "methods" and "hosts" options to security.firewalls

### DIFF
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -219,10 +219,10 @@ class SecurityServiceProvider implements ServiceProviderInterface, EventListener
                 $methods = isset($firewall['methods']) ? $firewall['methods'] : null;
 
                 unset(
-                    $firewall['pattern'], 
-                    $firewall['users'], 
-                    $firewall['security'], 
-                    $firewall['stateless'], 
+                    $firewall['pattern'],
+                    $firewall['users'],
+                    $firewall['security'],
+                    $firewall['stateless'],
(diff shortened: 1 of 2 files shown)


src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -344,14 +371,14 @@ public function register(Container $app)
                 if (is_string($rule[0])) {
                     $rule[0] = new RequestMatcher($rule[0]);
                 } elseif (is_array($rule[0])) {
-                    $rule[0] += [
+                    $rule[0] += array(
```
emirb: Please revert to short array syntax.

K0nias: can you please tell me why? its more readable my way.

stof: @emirb actually, all other code in Silex is using the long array syntax

emirb: @stof My bad, as I presumed standard would be short array syntax as we're supporting versions only above 5.4.
@K0nias disregard my comment then :)

K0nias: @emirb oh I misunderstand your comment. I thought that you were commenting associates array with string keys. Ok then. :)

GromNaN:
👍 you should add a test case.

K0nias:
@GromNaN added :)

fabpot:
Thank you @K0nias.

PR #1505, opened 2017-04-23
Use the simple-phpunit script in Travis and add PHP 7.1

skalpa:
This will fix the Travis build issues with PHP 7.

I upped the `symfony/phpunit-bridge` version requirement so we can use the `simple-phpunit` script that was added in 3.2.0, and updated the `script` config value.

While I was at it, I also added PHP 7.1 to the matrix (hopefully it's ok to do this in the same PR).


stof:
👍

skalpa:
@fabpot Apologies for pinging you, but it would be really great if you could merge this one as soon as you find a moment, as all the builds fail right now 😢 .

fabpot:
Thank you @skalpa.

PR #1506, opened 2017-04-23
E_USER_DEPRECATED: unicode asserts fix

rusranx:
E_USER_DEPRECATED: Using UTF-8 route requirements without setting the "utf8" option is deprecated since Symfony 3.2 and will throw a LogicException in 4.0. Turn on the "utf8" route option for variable in pattern in /vendor/symfony/routing/RouteCompiler.php on line 180

### DIFF
--- a/src/Silex/Route.php
+++ b/src/Silex/Route.php
@@ -68,8 +68,8 @@ class Route extends BaseRoute
     {
         $this->setRequirement($variable, $regexp);
 
-        if($isUnicode){
-            $this->setOption("utf8", true);
+        if ($isUnicode) {
+            $this->setOption('utf8', true);
         }
 
         return $this;


skalpa:
Doing that here seems completely arbitrary as this will only enable UTF-8 when a requirement is added. Instead, I'd suggest that you:
- Add a `routing.utf8` parameter that defaults to `false` to the `RoutingServiceProvider`
- Take the parameter value into account in the `route_factory`:
```php
    $app['routing.utf8'] = false;

    $app['route_factory'] = $app->factory(function ($app) {
        $route = new $app['route_class']();
        if ($app['routing.utf8']) {
            $route->setOption('utf8', true);
        }

        return $route;
    });
```
This will allow people to configure UTF-8 support globally and ensure every route created by Silex has the option set.

Also, you should probably add some tests to such a PR if you want it to get merged 😉.

rusranx:
Good idea, tnx

PR #1508, opened 2017-04-30
Added WebLink component integration

skalpa:
I made the necessary changes to support the new WebLink component:

- `HttpKernelServiceProvider` registers the `AddLinkHeaderListener` if the component is available
- `TwigServiceProvider` adds the `WebLinkExtension` if the component and the Twig bridge are available
- Both changes are tested
- The documentation has been amended (the link to the Symfony website doesn't work as 3.3 hasn't been released yet but I believe it's the good URL)


fabpot:
Thank you @skalpa.

PR #1510, opened 2017-05-02
Fix AssetServiceProviderTest::testGenerateAssetUrl

skalpa:
The test broke when symfony/symfony#22528 got merged. 

I added one check without the starting slash (the returned path should be relative to `base_path`) and modified the expectation of the check with the starting slash (the returned path should not be relative).

```
There was 1 failure:

1) Silex\Tests\Provider\AssetServiceProviderTest::testGenerateAssetUrl
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'/whatever-makes-sense/foo.css?css2'
+'/foo.css?css2'
```


skalpa:
I went too fast and the "fixed" test didn't work with 3.0 and 3.1.

Thus I just kept the test without the starting slash as the behavior of `getUrl()` in that case is the same from 2.7 to 3.3.

fabpot:
Thank you @skalpa.

PR #1511, opened 2017-05-02
Fix deprecation notice with symfony/twig-bridge 3.2+ in TwigServiceProvider

skalpa:
Fixes the deprecation triggered when instantiating `FormExtension` twig-bridge with 3.2+.

I check for the existence of `Symfony\Bridge\Twig\Extension\HttpKernelRuntime` to determine if we're using 3.2 as it was added in 3.2.0-beta1.

### DIFF
--- a/src/Silex/Provider/TwigServiceProvider.php
+++ b/src/Silex/Provider/TwigServiceProvider.php
@@ -109,7 +109,7 @@ class TwigServiceProvider implements ServiceProviderInterface
                         return new TwigRenderer($app['twig.form.engine'], $csrfTokenManager);
                     };
 
-                    $twig->addExtension(new FormExtension($app['twig.form.renderer']));
+                    $twig->addExtension(new FormExtension(class_exists(HttpKernelRuntime::class) ? null : $app['twig.form.renderer']));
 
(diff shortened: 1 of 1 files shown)


xabbuh:
The `FormExtension` class is part of the Twig bridge. Checking for the existence of a class from the Form component is not sufficient as the Form component in version 3.2 or higher may be used with an older version of the Twig bridge.

Also, you will then have to make sure that the Twig environment is passed to the Twig renderer engine (see the description of the deprecation in the extension class).

skalpa:
Good catch, thanks.

I changed the test and now check if `Symfony\Bridge\Twig\Extension\HttpKernelRuntime` exists (it was added by symfony/symfony@812fbb444f1d1e4cafff5876ce7b22d5b8d4adf8).

As for the renderer engine: I had already checked and the Twig environment is [already passed to the constructor](https://github.com/silexphp/Silex/blob/1b2f0394e28bb37cb3d86c84ba9a169b670c3fe9/src/Silex/Provider/TwigServiceProvider.php#L103).

fabpot:
Thank you @skalpa.

PR #1512, opened 2017-05-02
Fix duck type in MiddlewareListener constructor

skalpa:
The typehint in the MiddlewareListener constructor should be `Container` and not `Application` (must be a leftover from Silex 1.x).

It's instantiated from `HttpKernelServiceProvider::subscribe(Container $app, EventDispatcherInterface $dispatcher)`.

(The failures are unrelated, due to #1510)

### DIFF
--- a/src/Silex/EventListener/MiddlewareListener.php
+++ b/src/Silex/EventListener/MiddlewareListener.php
@@ -11,12 +11,12 @@
 
 namespace Silex\EventListener;
 
+use Pimple\Container;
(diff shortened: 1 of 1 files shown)


fabpot:
`$this->app` is passed to middleware, have you checked that those expect Container instances as well?

skalpa:
That's a very good point.

And it actually makes sense to pass the application to middlewares. I guess I didn't think this one through (I shouldn't push PRs late at night).

Sorry for the noise.

PR #1514, opened 2017-05-09
Fixed WebTestCase compatibility with PHPUnit 6+

skalpa:
Makes `WebTestCase` inherit from `PHPUnit\Framework\TestCase` instead of `PHPUnit_Framework_TestCase`.

Strictly speaking it's a BC break, but as the namespaced class is now available in PHPUnit 4.8.35, 5.4+ and 6.x it's probable safe.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,6 +50,9 @@
         "monolog/monolog": "^1.4.1",
         "symfony/web-link": "^3.3"
     },
+    "conflict": {
+        "phpunit/phpunit": "<4.8.35 || >= 5.0, <5.4.3"
+    },
     "replace": {
         "silex/api": "self.version",
         "silex/providers": "self.version"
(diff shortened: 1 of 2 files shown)


stof:
you can mitigate the BC break for people installing PHPunit as a dev requirement in their project by adding this conflict rule in composer.json (which will force composer to select a compatible PHPUnit version):

```json
{
    "conflict": {
        "phpunit/phpunit": "<4.8.35 || >= 5.0, <5.4.3"
    }
}
```

People installing PHPUnit separately will still need to deal with supported versions themselves (but anyway, incompatible versions are EOLed versions)

fabpot:
I like @stof suggestion.

stof:
Note that this is already what we do in `symfony/phpunit-bridge` in 3.3 (we should also do it in HttpKernel in releases where the WebTestCase was migrated though, but we have not done it yet)

skalpa:
Here you go, I added the `conflict` section to `composer.json` :-)

fabpot:
Thank you @skalpa.

PR #1515, opened 2017-05-09
Use PHPUnit\Framework\TestCase in Silex tests

skalpa:
I changed all the tests so they use the namespaced class instead of `PHPUnit_Framework_TestCase`.

That doesn't really bring anything, but will be one less thing to change when Silex switches to PHP7/PHPUnit 6.


fabpot:
Thank you @skalpa.

lukaszgoworko:
nice 👍 thank you

PR #1517, opened 2017-05-14
Clarify documentations around protected closures.

jordanlev:
The docs state that "protected closures do not get access to the container", but this is somewhat misleading (it confused me for a quite some time). It's true that protected closures are not passed the container as a function argument, but they can access the container via `use ($app)`.


fabpot:
Thank you @jordanlev.

fabpot:
I've also added an example in 694fe55 to make it extra clear.

PR #1518, opened 2017-05-15
add json manifest version strategy support

MatTheCat:


fabpot:
Thank you @MatTheCat.

PR #1521, opened 2017-05-18
Update web_servers.rst

progsmile:
Reformat code listing

---
This change is [Reviewable](https://reviewable.io/reviews/silexphp/silex/1521)


SpacePossum:
:+1:  for fixing this (https://silex.sensiolabs.org/doc/2.0/web_servers.html looks off)
however it would be nice to remove the trailing spaces from the PR I think

fabpot:
Thank you @progsmile.

PR #1522, opened 2017-05-19
Error using EsiFragment with provider and twig functions

mihailov-vf:
Trying to use twig functions "controller" and "render_esi" like [Symfony esi reference](http://symfony.com/doc/current/http_cache/esi.html) , i've noticed the following error:
An exception has been thrown during the rendering of a template ("You must use a URI when using the ESI rendering strategy or set a URL signer.").

It's the lack of a UriSigner on the constructor.


fabpot:
Thank you @mihailov-vf.

PR #1524, opened 2017-05-24
Add symfony/cache service provider

skalpa:
I tried to keep it as close to the framework implementation as possible:

- The `system` pool uses `AbstractAdapter::createSystemCache()` 
  (eg: ApcuAdapter, PhpFilesAdapter or FilesystemAdapter depending on what's available)
- The `app` pool uses Redis/Memcached if a DSN has been specified, or FilesystemAdapter
  otherwise
- Users can create additional pools based on `system` or `app` by adding entries to
  `cache.pools.options`

The documentation is missing, but I'd like to get the implementation reviewed first.

### Minimum configuration
```php
$app->register(new CacheServiceProvider(), [
    'cache.directory' => '/path/to/cache',
]);

$pool = $app['cache.pools']['system'];
$item = $pool->getItem('key');
```

### Example usage
```php

$app->register(new CacheServiceProvider(), [
    'cache.directory' => '/path/to/cache',
    'cache.namespace_prefix' => 'prepended_to_namespaces_',

    'cache.system.namespace' => 'system',
    'cache.system.default_lifetime' => 0,
    'cache.system.version' => 'apcu-version-string',

    'cache.app.namespace' => 'app',
    'cache.app.default_lifetime' => 0,
    'cache.app.dsn' => 'redis://localhost',
    'cache.app.connection_options' => ['timeout' => 3],

    'cache.pools.option' => [
        // Build namespace using cache.system.namespace + pool name, use default_lifetime
        'system-based' => 'system',
        // Build namespace using cache.app.namespace + pool name, use default_lifetime
        'app-based' => 'app',
        // Use custom namespace and lifetime
        'custom' => [
            'adapter' => 'app',
            'namespace' => 'custom_ns',
            'lifetime' => 42,
        ],
    ],
]);

// You can override the factories to customize how all the pools of a specific type are created

$app['cache.system.factory'] = $app->protect(function($namespace, $lifetime = null) {
    return new NullAdapter();
});

// Add a custom pool not based on the default system or app adapters

$app->extend('cache.pools', function ($pools) use ($app) {
    $pools['db-cache'] = function () use ($app) {
        return new PdoAdapter('dsn', $app['cache.namespace_prefix'].'foo', 42);
    };

    return $pools;
});

$pool = $app['cache.pools']['system'];
$item = $pool->getItem('key');
```

### About the tests

Instead of adding one environment with APCu, one with Redis, etc... I just 
added one will all the services enabled. In my opinion that should be 
enough to cover all the cases as I'm not trying to test the cache component 
but only that the provider passes the correct parameters to the classes 
it instantiates.

I don't test lifetimes as it's slow... Now I could but that would mean
adding a bunch of `sleep()` calls to the tests.

Note:
- I merged #1523 into this PR to be able to test against 3.3/3.4
- A lot of the tests code (specifically the setUp/tearDown) is directly based on Nicolas' work on `symfony/cache`, so he'll deserve as much (if not more) credit on this one as I barely ported what he had done

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,18 +9,31 @@ env:
 cache:
     directories:
       - $HOME/.composer/cache/files
+      - .phpunit
 
 before_install:
     - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then phpenv config-rm xdebug.ini; fi
 
 before_script:
-    # symfony/*
-    - sh -c "if [ '$TWIG_VERSION' != '2.0' ]; then sed -i 's/~1.8|~2.0/~1.8/g' composer.json; composer update; fi"
-    - sh -c "if [ '$SYMFONY_DEPS_VERSION' = '3.0' ]; then sed -i 's/~2\.8|^3\.0/3.0.*@dev/g' composer.json; composer update; fi"
-    - sh -c "if [ '$SYMFONY_DEPS_VERSION' = '3.1' ]; then sed -i 's/~2\.8|^3\.0/3.1.*@dev/g' composer.json; composer update; fi"
-    - sh -c "if [ '$SYMFONY_DEPS_VERSION' = '3.2' ]; then sed -i 's/~2\.8|^3\.0/3.2.*@dev/g' composer.json; composer update; fi"
-    - sh -c "if [ '$SYMFONY_DEPS_VERSION' = '' ]; then sed -i 's/~2\.8|^3\.0/2.8.*@dev/g' composer.json; composer update; fi"
-    - composer install
+    # Twig 1.x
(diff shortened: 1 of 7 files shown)


hkdobrev:
@skalpa Do you think you could resurrect this one? 

- [ ] Fix conflicts after #1523 was merged to leave things here only relevant to the cache.
- [ ] Check if there aren't any updated to `symfony/cache` that need to be reflected here.
- [ ] Add a documentation page for the provider with the example usage

I'm not much familiar with the `symfony/cache` component as I'm used to another PSR-6 implementation to do it myself, so if not you, someone else familiar with it would be more suitable to bring this up to date so nothing important is missed.

fabpot:

PR #1533, opened 2017-06-12
Remove outdated example code from documentation

jnvsor:
The `user` method hasn't been part of the security
trait since 7052d97 over 2 years ago.


fabpot:
Good catch, thanks @jnvsor.

PR #1534, opened 2017-06-12
RedirectableUrlMatcher needs to return a proper array with the _route parameter

slipo:
Symfony's HttpUtils's checkRequestPath() expects RedirectableUrlMatcher to return an array with '_route'. This throws a notice `Undefined index: _route`:

```
        $coll = new RouteCollection();
        $coll->add('foo', new Route('/foo', array(), array(), array(), '', array('https')));
        $matcher = new RedirectableUrlMatcher($coll, new RequestContext());
        $httpUtils = new HttpUtils(null, $matcher);
        $request = Request::create('http://example.com/foo');
        $httpUtils->checkRequestPath($request, 'foo');
```

It's an issue with my app although I haven't yet taken the time to make a simplified app to show the issue besides the above.


fabpot:
Thank you @slipo.

PR #1536, opened 2017-07-13
Added PSR-11 service provider

skalpa:
PSR-11 service provider that adds:
- A `psr11` service
- Argument value resolvers that inject the PSR-11 container into controllers:

```php
use Psr\Container\ContainerInterface;

$app->get('/', function (ContainerInterface $container) {
    $service = $container->get('service');
});
```

The resolution is performed by an `ArgumentValueResolverInterface` if the kernel is recent enough, or by a custom `ControllerResolverInterface` that decorates the `resolver` service otherwise (so we don't break BC).

AFAIK it's the only thing Silex will need, but if somebody can think of another use for the PSR-11 container give me a shout.

I'll add the documentation once Pimple has been updated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     ],
     "require": {
         "php": ">=5.5.9",
-        "pimple/pimple": "^3.2",
+        "pimple/pimple": "^3.2.1",
         "symfony/event-dispatcher": "~2.8|^3.0",
         "symfony/http-foundation": "~2.8|^3.0",
         "symfony/http-kernel": "~2.8|^3.0",
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     ],
     "require": {
         "php": ">=5.5.9",
-        "pimple/pimple": "~3.0",
+        "pimple/pimple": "^3.2",
         "symfony/event-dispatcher": "~2.8|^3.0",
         "symfony/http-foundation": "~2.8|^3.0",
         "symfony/http-kernel": "~2.8|^3.0",
--- a/doc/providers/psr11.rst
+++ b/doc/providers/psr11.rst
@@ -4,10 +4,10 @@ PSR-11
 The *Psr11ServiceProvider* provides a container for accessing your application
 services using the PSR-11 ``ContainerInterface`` interface.
 
-PSR-11 is a standard that describes a common interface for dependency injection
+`PSR-11`_ is a standard that describes a common interface for dependency injection
 containers. Using the PSR-11 container will allow you to:
 
-* Use objects that expect a ``ContainerInterface`` instance.
+* Use objects that expect a ``Psr\Container\ContainerInterface`` instance.
 
 * Decouple your own code from the ``Silex\Application`` class (something that
   could prove useful if you intend to port your application to the Symfony
@@ -16,8 +16,7 @@ containers. Using the PSR-11 container will allow you to:
 Services
 --------
 
-* **container**: A container that implements `ContainerInterface
-  <https://github.com/container-interop/fig-standards/blob/master/proposed/container.md>`_
+* **container**: A container that implements ``Psr\Container\ContainerInterface``
   and give you access to all your services.
 
   Example usage::
@@ -25,8 +24,7 @@ Services
     $container = $app['container'];
     $service = $container->get('service');
 
-* **service_locator.factory**: A factory that creates `PSR-11 service locators
-  <https://github.com/silexphp/Pimple/blob/master/README.rst#using-the-psr-11-servicelocator>`_.
+* **service_locator.factory**: A factory that creates `PSR-11 service locators`_.
 
   Example usage::
 
@@ -103,3 +101,6 @@ instead of the whole container by using the ``convert()`` function:
     })->convert('container', function () use ($app) {
         return $app['service_locator.factory'](array('foo', 'bar'));
     });
+
+.. _PSR-11: https://github.com/container-interop/fig-standards/blob/master/proposed/container.md
+.. _PSR-11 service locators: https://github.com/silexphp/Pimple/blob/master/README.rst#using-the-psr-11-servicelocator
--- a/doc/providers/psr11.rst
+++ b/doc/providers/psr11.rst
@@ -1,0 +1,105 @@
+PSR-11
+======
+
+The *Psr11ServiceProvider* provides a container for accessing your application
+services using the PSR-11 ``ContainerInterface`` interface.
+
+PSR-11 is a standard that describes a common interface for dependency injection
+containers. Using the PSR-11 container will allow you to:
+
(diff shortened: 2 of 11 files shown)


src/Silex/Provider/Psr11ServiceProvider.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+class Psr11ServiceProvider implements ServiceProviderInterface
+{
+    public function register(Container $app)
+    {
+        $app['psr11'] = function ($app) {
+            return new PsrContainer($app);
+        };
+
+        if (Kernel::VERSION_ID >= 30100) {
```
fabpot: version looks weird to me, missing one digit.

fabpot (changes requested):

src/Silex/Provider/Psr11ServiceProvider.php
skalpa: Mmmhh... It might feel weird but then that would be [your fault](https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/HttpKernel/Kernel.php#L64) 😉. 

The `VERSION_ID` is XYYZZ, so to test if we have at least 3.1.0 it should be greater than 30100. That's how it's done in [other places in Silex](https://github.com/silexphp/Silex/blob/master/src/Silex/Provider/HttpKernelServiceProvider.php#L45).

src/Silex/Provider/Psr11ServiceProvider.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Provides a PSR-11 container and argument value resolver.
+ *
+ * @author Pascal Luna <skalpa@zetareticuli.org>
+ */
+class Psr11ServiceProvider implements ServiceProviderInterface
+{
+    public function register(Container $app)
+    {
+        $app['psr11'] = function ($app) {
```
GromNaN: psr11 is a very unclear name for beginners and people ignoring php-fig.

"container" or "service_locator" would be more explicit on the purpose of this service.

GromNaN (changes requested):

skalpa:
I added a `service_locator.factory` to create instances of `Pimple\Psr11\ServiceLocator` and the documentation. 

I'll check everything again once silexphp/Pimple#233 and silexphp/Pimple#236 have been merged/released.

@GromNaN I renamed the container service to `container` as I agree the previous name wasn't a great choice. Now, I don't see why somebody who has never heard of the standard would want to use this provider, so I honestly hesitated between this and `psr11.container`... Still, I kept `container` for now as it was the most simple option / the name you used in your own provider.

GromNaN (approved):

composer.json
```diff
@@ -16,7 +16,7 @@
     ],
     "require": {
         "php": ">=5.5.9",
-        "pimple/pimple": "~3.0",
+        "pimple/pimple": "^3.2",
```
fabpot: We could probably switch to `^3.2.1`

doc/providers/psr11.rst
```diff
@@ -0,0 +1,105 @@
+PSR-11
+======
+
+The *Psr11ServiceProvider* provides a container for accessing your application
+services using the PSR-11 ``ContainerInterface`` interface.
+
+PSR-11 is a standard that describes a common interface for dependency injection
```
fabpot: You could add a link to the PSR on the FIG website.

doc/providers/psr11.rst
```diff
@@ -0,0 +1,105 @@
...
+
+* Decouple your own code from the ``Silex\Application`` class (something that
+  could prove useful if you intend to port your application to the Symfony
+  full-stack framework in the future).
+
+Services
+--------
+
+* **container**: A container that implements `ContainerInterface
+  <https://github.com/container-interop/fig-standards/blob/master/proposed/container.md>`_
```
fabpot: URLs should be referenced as footnotes (I think that's what we are doing elsewhere)

skalpa:
@fabpot I upped the Pimple version to 3.2.1 and addressed your two comments

derrabus (approved):

derrabus:
For applications that want to migrate away from Silex, this feature would actually be really nice to have.

xabbuh (changes requested):
not all build failures look unrelated to the changes made here (see https://travis-ci.org/silexphp/Silex/jobs/255495935#L382)

skalpa:
@xabbuh Are you sure it would make sense to finish this ? From what I understand, @fabpot already switched the project to _maintenance only_ and that is a new feature.

I've been more than busy with other things lately, so if someone can confirm that this PR would get merged if I rebase and fix it I could try and find some time to do it, but would like to avoid wasting time on something that will never land.

fabpot:
I think being PSR-11 is a nice feature to have before the end of life as it would ease the move from Silex to something else.

stof:
There is https://github.com/derrabus/silex-psr11-provider providing exactly this feature already (but without support for Symfony < 3.1, while this PR does, but I don't think this is an issue as the typical migration would be to go to Flex, and so to Symfony 3.4+)

fabpot:
Then, I would says that we don't need to spend more time here.

hkdobrev:
I guess we could add a cookbook entry to the docs mentioning the PSR-11 container and the https://github.com/derrabus/silex-psr11-provider provider.

fabpot:
@hkdobrev Adding some docs about the third party provider is indeed a great idea and that would fix this issue.

fabpot:
Now, if Silex has a future, it might be interesting to actually provide PSR-11 out of the box. That's really up to the new team, but I think it would make sense.

derrabus:
I've created the `derrabus/silex-psr11-provider` package from a code snippet that I've been using in some Silex to Symfony migrations. I needed that functionality for my [blog post](https://medium.com/@derrabus/decoupling-silex-applications-from-the-service-container-implementation-635263f05e52) and, because of the announced EOL of Silex, I did not expect this PR to be merged.

My package only provides the PSR-11 container as a service and the argument resolver. In addition, this PR also provides Symfony 3.0 compatibility and service locators. So, in a way, my package is a minimalist version of this PR.

I agree that it would make perfect sense to have PSR-11 support in Silex core. And I would offer to deprecate my package and update my blog post once there is a stable Silex 2 release with a PSR-11 container.

I _could_ also offer to submit my package as an alternative PR, but this PR is @skalpa's baby and I don't want to oust him on this one.

hkdobrev (approved):
Should this be merged or just mention the 3rd-party solution in the docs?

I also prefer having a solution in the core and announced in the changelog so more people could migrate to PSR-11.

fabpot:
I still think that it makes sense to have built-in PSR-11 support if Silex is maintained in the long term. If not, it will become irrelevant at some point.

fabpot:

PR #1539, opened 2017-07-20
Handle the ControllerResolverInterface::getArguments deprecation in tests

skalpa:
Fixes deprecation notices/failures due to the deprecation of `ControllerResolverInterface::getArguments()`:

1. `Silex\ControllerResolver` [is not used by Silex](https://github.com/silexphp/Silex/blob/master/src/Silex/Provider/HttpKernelServiceProvider.php#L37) when using HttpKernel >= 3.1.0, so I skipped the only test present in `ControllerResolverTest` with recent versions, as it failed when testing against HttpKernel v4  (master).

2. `ServiceControllerResolverTest::testShouldDelegateGetArguments()` only asserts that the deprecated method is called, so added a `@group legacy` to handle the notice, and skipped the test with v4 to prevent a warning (PHPUnit complains when you mock a non-existent method).

### DIFF
--- a/tests/Silex/Tests/ControllerResolverTest.php
+++ b/tests/Silex/Tests/ControllerResolverTest.php
@@ -15,6 +15,7 @@ use PHPUnit\Framework\TestCase;
 use Silex\ControllerResolver;
 use Silex\Application;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @skalpa.

PR #1543, opened 2017-08-17
The general case is that 'app' may be located anywhere, accessible through an Alias.

PABourdin:


fabpot:
Thank you @PABourdin.

PR #1546, opened 2017-08-25
Add undocumented available parameter: base_path

Redominus:


fabpot:
Thank you @Redominus.

PR #1549, opened 2017-08-27
Fix wrong namespace forward/backward slash

dol:


fabpot:
Good catch, thanks @dol.

dol:
Catched by being lazy. Copy/pasting off the website and wondering why it's not working. 😉

PR #1550, opened 2017-09-21
Register UserPasswordValidator in SecurityServiceProvider boot method

### LINKED ISSUE #1531: Remove isset in SecurityServiceProvider to prevent dependency on ValidatorServiceProvider?
In the [SecurityServiceProvider](https://github.com/silexphp/Silex/blob/6f539e087f1178cfce6579eb7a080c4fa3805f38/src/Silex/Provider/SecurityServiceProvider.php#L661) there's an `if` statement that requires the `validator` to exist.

As long as that if is there registering validator after security will cause a nice fat fatal error when using the `UserPassword` constraint, and I can't imagine it's there for performance reasons.

Is that if even neccesary?

MatTheCat:
The current implementation requires to register the `ValidatorServiceProvider` **before** the `SecurityServiceProvider` in order to use the `UserPasswordValidator`. Moving its registration inside the `boot` method allow to register them in any order and fixes #1531.

### DIFF
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -657,14 +657,6 @@ class SecurityServiceProvider implements ServiceProviderInterface, EventListener
                 return new AnonymousAuthenticationProvider($name);
             };
         });
-
-        if (isset($app['validator'])) {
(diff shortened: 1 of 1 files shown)


fabpot:
Doing registration during `boot` is conceptually wrong. #1606 is a "better" alternative. Thanks.

PR #1551, opened 2017-09-22
[RFC][WIP] Allow separate logging channels

Xymanek:
Right now using default providers has a big disadvantage: all logs (database, security, etc) are dumped into app channel. This makes the logs harder to read and harder to manage (eg. I want security logs in one file but rest in another). Currently I see two ways to improve this:

### Prototype-based
Introduce a logger factory (eg `monolog.logger._proto`) which will be responsible for creating loggers with a given name/channel and use it in other providers (if present). Closely mimics MonologBundle's behaviour but isn't BC in case somebody extended the `monolog` service (I know my code will break)

 ### `\Monolog\Logger::withName()`
Looks to be a safer bet BC-wise however (due to my lack of knowledge of monolog's inner workings) I'm not sure if this can cause any potential problems in future

When introducing this it's also important to think about other libraries out there (starting with the [profiler](https://github.com/silexphp/Silex-WebProfiler/blob/master/WebProfilerServiceProvider.php#L70))

Looking forward to any comments and suggestions

P.S. I know about https://silex.symfony.com/doc/2.0/cookbook/multiple_loggers.html but I'm talking about services where `$app['logger']` is hard coded

P.P.S. `MonologServiceProvider` is so opinionated and inflexible that the above tutorial doesn't even use it. Maybe it's better to try and make it more flexible first? Although I have no idea how to do that in BC-way

### DIFF
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -65,6 +65,9 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
         $this['debug'] = false;
         $this['charset'] = 'UTF-8';
         $this['logger'] = null;
+        $this['logger.request'] = function () {
+            return $this['logger'];
+        };
 
         $this->register(new HttpKernelServiceProvider());
         $this->register(new RoutingServiceProvider());
--- a/src/Silex/Application.php
+++ b/src/Silex/Application.php
@@ -11,6 +11,7 @@
 
 namespace Silex;
 
+use Monolog\Logger;
 use Pimple\Container;
 use Pimple\ServiceProviderInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
@@ -66,7 +67,13 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
         $this['charset'] = 'UTF-8';
         $this['logger'] = null;
         $this['logger.request'] = function () {
-            return $this['logger'];
+            $logger = $this['logger'];
+
+            if ($logger instanceof Logger) {
+                $logger = $logger->withName('request');
+            }
+
+            return $logger;
         };
 
         $this->register(new HttpKernelServiceProvider());
--- a/src/Silex/Provider/RoutingServiceProvider.php
+++ b/src/Silex/Provider/RoutingServiceProvider.php
@@ -76,7 +76,12 @@ class RoutingServiceProvider implements ServiceProviderInterface, EventListenerP
                 return $app['request_matcher'];
             });
 
-            return new RouterListener($urlMatcher, $app['request_stack'], $app['request_context'], $app['logger']);
+            return new RouterListener(
+                $urlMatcher,
+                $app['request_stack'],
+                $app['request_context'],
+                $app['logger.request']
+            );
         };
     }
 
--- a/src/Silex/Provider/DoctrineServiceProvider.php
+++ b/src/Silex/Provider/DoctrineServiceProvider.php
@@ -11,6 +11,7 @@
 
 namespace Silex\Provider;
 
+use Monolog\Logger;
 use Pimple\Container;
 use Pimple\ServiceProviderInterface;
 use Doctrine\DBAL\DriverManager;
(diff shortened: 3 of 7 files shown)


src/Silex/Application.php
```diff
@@ -65,6 +66,15 @@ public function __construct(array $values = array())
...
         $this['logger'] = null;
+        $this['logger.request'] = function () {
+            $logger = $this['logger'];
+
+            if ($logger instanceof Logger) {
+                $logger = $logger->withName('request');
+            }
+
+            return $logger;
+        };
```
hkdobrev: I'd move the registration of `'logger.request'` to the `HttpKernelServiceProvider` instead.

All usages of `$app['logger.request']` could be replaced with `$app['logger.request'] ?? $app['logger']`.

src/Silex/Provider/RoutingServiceProvider.php
```diff
@@ -76,7 +76,12 @@ public function register(Container $app)
                 return $app['request_matcher'];
             });
 
-            return new RouterListener($urlMatcher, $app['request_stack'], $app['request_context'], $app['logger']);
+            return new RouterListener(
+                $urlMatcher,
+                $app['request_stack'],
+                $app['request_context'],
+                $app['logger.request']
```
hkdobrev: I'd recommend to have a different logger - `logger.router` for the route matching. Not only, because this is not about the request, but also, because route matches are being logged on the `INFO` level in contrast to other loggers logging on the `DEBUG` level and users might want to do something different only with `logger.router` like changing the log level threshold.

hkdobrev (changes requested):

hkdobrev:
Do you think you'd be able to fix the merge conflicts? Thank you! 🚀

Xymanek:
Oh wow, I gave up on this PR quite a while ago, lol

Sure, I will fix it next week :)

hkdobrev:
On a separate note, regarding your comment on the multiple loggers example, I've used it like that:

```php
use Pimple\Container;
use Psr\Log\LogLevel;
use Monolog\Formatter\LineFormatter;
use Monolog\Handler\SlackHandler;
use Silex\Provider\MonologServiceProvider;

$app->register(new MonologServiceProvider(), [
	'monolog.logfile' => '/foo/bar/logfile,
    'monolog.level' => $app['debug'] ? LogLevel::DEBUG : LogLevel::WARNING,
	'logger.slack_updates' => function (Container $app) {
		$logger = new $app['monolog.logger.class']('slack_updates');

        $slackHandler = new SlackHandler(
            getenv('SLACK_TOKEN'),
            '#updates',
            'My awesome app',
            false,
            'ghost',
            LogLevel::INFO
        );
        $slackHandler->setFormatter(new LineFormatter('%message%'));
        $logger->pushHandler($slackHandler);

        return $logger;
	}),
]);
```

If you think this is a more sensible example, we could add it to the docs.

Xymanek:
The problem with your approach is that it doesn't allow things like debug processor to hook into your logger. In `MonologBundle` this can be done by adding method calls to `monolog.logger_prototype` abstract service. I would like to introduce such functionality for silex as well.

## My original idea
Use a factory closure. Here is a simple demo to give you an idea how that is supposed to work:

```php
class MyClass {
    public function __construct (string $name)
    {
        $this->println('Created object with name: ' . $name);
    }

    public function println (string $str)
    {
        echo $str . PHP_EOL;
    }
}

$container = new Pimple\Container();

// Original definition (eg. MonologServiceProvider)
$container['factory'] = $container->protect(function (string $name) {
    $obj = new MyClass($name);
    $obj->println('Original');
    
    return $obj;
});
$container['obj'] = function ($c) {
    return $c['factory']('test');
};


// Augmenting (eg. debug)
$factory = $container['factory'];
$container['factory'] = $container->protect(function (string $name) use ($factory) {
    $obj = $factory($name);
    $obj->println('Extend');

    return $obj;
});

// Userland code
$container['obj']->println('Body');
```

Result:
```
Created object with name: test
Original
Extend
Body
```

[Implementation](https://github.com/Xymanek/Silex/blob/70a403c66d130b41d48857c8492924ac0f864240/src/Silex/Provider/MonologServiceProvider.php#L56-L75) in `MonologServiceProvider`.

## My second idea
Seeing how convoluted the original method was I decided to use `Monolog\Logger::withName()` to simply duplicate loggers while retaining "configuration". In this approach `Pimple\Container::extend('monolog')` is enough to augment the logger with additional handlers, processors, etc and consumers simply need get `logger` and then duplicate it.

The only downside of second approach is that there is no difference between app logger (`$app['logger]`) and the "prototype". This means that any changes done to app logger will be reflected in all other loggers. Unfortunately I have no idea how to improve this while keeping BC (where all providers used the app logger)

Let me know which approach I should use when fixing this PR

Xymanek:
@hkdobrev ping

fabpot:

PR #1552, opened 2017-09-25
Explicitly add `symfony/debug` as a requirement

h-bragg:
The code: https://github.com/silexphp/Silex/blob/master/src/Silex/ExceptionHandler.php#L37 uses `symfony/debug` classes directly. However it is currently only installed by inference from `symfony/http-kernel` when not installing the development dependencies.

If you do a `composer update --prefer-stable --prefer-lowest` from outside of silex it will grab the `2.6` version of `symfony/debug` which does not include the `getHtml` method on `DebugExceptionHandler` and thus everything fails.

This was discovered in: graze/silex-trailing-slash-handler#9

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     "require": {
         "php": ">=5.5.9",
         "pimple/pimple": "~3.0",
+        "symfony/debug": "~2.8|^3.0",
         "symfony/event-dispatcher": "~2.8|^3.0",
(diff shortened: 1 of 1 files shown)


fabpot:
I've just tested and (today) it installs `symfony/debug` 2.8.0. So, upstream was probably fixed at some point.

PR #1556, opened 2017-10-04
update session_storage.rst

adouiri:
remove extra spaces


fabpot:
Good catch, thanks @adouiri.

PR #1558, opened 2017-10-09
Update guard_authentication.rst

symac:
I am still working on it to get authentication working in my project so my suggestion might be false, but I believe a ```use``` is missing in the header of this example.


fabpot:
Thank you @symac.

Issue #1566, opened 2017-11-08
Tes failure with PHP 7.2


remicollet:
From Fedora QA
https://apps.fedoraproject.org/koschei/groups/php?order_by=state%2C-started&collection=f28

Reproduced with master
```
There was 1 error:

1) Silex\Tests\Provider\SessionServiceProviderTest::testRegister
ini_set(): Headers already sent. You cannot change the session module's ini settings at this time

/tmp/Silex/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:105
/tmp/Silex/vendor/symfony/http-foundation/Session/Storage/Handler/NativeFileSessionHandler.php:52
/tmp/Silex/src/Silex/Provider/SessionServiceProvider.php:49
/tmp/Silex/vendor/pimple/pimple/src/Pimple/Container.php:118
/tmp/Silex/src/Silex/Provider/SessionServiceProvider.php:55
/tmp/Silex/vendor/pimple/pimple/src/Pimple/Container.php:118
/tmp/Silex/tests/Silex/Tests/Provider/SessionServiceProviderTest.php:33
```

lracicot:
I still have the same issue with `http-foundation`
```
name     : symfony/http-foundation
versions : * v3.4.3
``` 

Whenever I run phpunit, I get: 
```
ini_set(): Headers already sent. You cannot change the session module\'s ini settings at this time', '~/project/app/vendor/symfony/http-foundation/Session/Storage/Handler/NativeFileSessionHandler.php (line 52)
```

GBeushausen:
What can we do to make the tests work with 7.2? I'm not quite sure what the status is. The code in #1567 seems to be doing the trick. Why was the PR canceled?

SpacePossum:
IIRC I tried to do to much at the same time in one PR ;)
Dropping the CS, SCA and modernize changes leaves the main issue:
finding a way to run the test suit on PHP5.3 up to and including PHP7.2, given the changes done within PHPUnit there is not always an easy path. TBH I stopped working on the PR as around that time Silex was reported to become EOL. IMHO best/easiest way would be start again with a clean up to date PR.

GBeushausen:
@SpacePossum thanks. I'll try to work on that. From what i see this issue is one of the main show stoppers to the further development of Silex.

[...]

PR #1567, opened 2017-11-08
PHP7.2 compat

SpacePossum:
`symfony/validator": "~2.3|3.0.*` is not compat with PHP7.2, 
this has been resolved on a higher minor so lets drop that constraint.
I'll check the failing session test later when Travis has a meaningful report here.

closes https://github.com/silexphp/Silex/issues/1566

when this is all OK I'll prep. one for `master` line

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,12 +53,11 @@
         "psr-4": { "Silex\\Tests\\" : "tests/Silex/Tests" }
     },
     "config": {
-        "optimize-autoloader": true,
         "sort-packages": true
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.3.x-dev"
+            "dev-master": "1.4.x-dev"
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -23,33 +23,40 @@
         "symfony/routing": "~2.3|3.0.*"
     },
     "require-dev": {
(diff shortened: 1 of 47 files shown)


SpacePossum:
Maybe best to throw PHPUnit as a `dev-dep` in `composer.json` as well?
It would save a lot of configuration tweaking for `travis`.

composer.json
```diff
@@ -23,38 +23,42 @@
...
         "psr-4": { "Silex\\Tests\\" : "tests/Silex/Tests" }
     },
+    "config": {
+        "optimize-autoloader": true,
+        "sort-packages": true
+    },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.3.x-dev"
+            "dev-master": "2.3.x-dev"
```
stof: does not make any sense, as you commit this in 1.3

composer.json
```diff
@@ -23,38 +23,42 @@
...
+        "twig/twig": "~1.28|~2.0"
     },
     "autoload": {
         "psr-4": { "Silex\\": "src/Silex" }
     },
     "autoload-dev" : {
         "psr-4": { "Silex\\Tests\\" : "tests/Silex/Tests" }
     },
+    "config": {
+        "optimize-autoloader": true,
```
stof: this has no reason to be here

PR #1570, opened 2017-11-15
Update security.rst - security context sharing

Flyrell:
I'd like to update the security.rst as I found the lack of information about the security context sharing between multiple firewalls. I know a lot of people are having problems not knowing this, so it would be nice to mention it. Please, check the correctness and approve it if you find it useful. Thanks.


Flyrell:
@chihiro-adachi You could've at least explain why you closed it. Maybe I'm not aware of something here.

Flyrell:
@fabpot any chance this will be merged?

fabpot:
Thank you @Flyrell.

PR #1580, opened 2017-12-10
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertFalse` instead of strict comparison with `false` keyword;
- `assertArrayHasKey` instead of `isset` function.


fabpot:
Thank you @carusogabriel.

Issue #1582, opened 2017-12-14
Swiftmailer SSL stream_context_options cannot be set


leClaude:
The stream_context_options cannot be set using the main swiftmailer provider.

hkdobrev:
This is now resolved with #1583.

PR #1583, opened 2017-12-14
Swiftmailer SSL stream_context_options cannot be set #1582 :

leClaude:
Adding stream_context_options to the $app['swiftmailer.options']


fabpot:
Thank you @leClaude.

PR #1594, opened 2018-02-13
Do not ignore swiftmailer.options if present.

lgnstn:
Allow configuring Swiftmailer SMTP as stated here https://silex.symfony.com/doc/2.0/providers/swiftmailer.html#parameters


srsbiz:
See https://github.com/silexphp/Silex/issues/1486

Issue #1591, opened 2018-01-15
"Welcome!" page with $app['debug'] = true


ProgMiner:
Is it right, "Welcome!" page opens with `$app['debug'] = true`?
I think in `RoutingServiceProvider` file in `RouterListener` must be also a `$app['debug']` parameter.
Sorry if it isn't.

https://github.com/silexphp/Silex/blob/e970d12a5b5a203fbd859525ce406b22e73f9f61/src/Silex/Provider/RoutingServiceProvider.php#L79

flip111:
> I think in RoutingServiceProvider file in RouterListener must be also a $app['debug'] parameter.

why do you think so? can you be more specific here ?

ProgMiner:
Because a Welcome page [contains](https://github.com/symfony/symfony/blob/b0facfec018093fb9ad65658f654ab7af06f6f87/src/Symfony/Component/HttpKernel/Resources/welcome.html.php#L79) that:
```
<div id="comment">
    <p>
        You're seeing this message because you have debug mode enabled and you haven't configured any URLs.
    </p>
</div>
```

flip111:
What does that piece of html have to do with the RoutingServiceProvider ?

ProgMiner:
Maybe I don't understand something, but I think this is obviously.
A `RouterListener` from the Symfony must sends Welcome page **only** when application in debug-mode.
Possibly I mistaked by sending a quote from HTML source. It's [from](https://github.com/symfony/symfony/blob/b0facfec018093fb9ad65658f654ab7af06f6f87/src/Symfony/Component/HttpKernel/EventListener/RouterListener.php#L133) `RouterListener`:
```
} catch (ResourceNotFoundException $e) {
    if ($this->debug && $e instanceof NoConfigurationException) {
        $event->setResponse($this->createWelcomeResponse());

        return;
    }

    $message = sprintf('No route found for "%s %s"', $request->getMethod(), $request->getPathInfo());

    if ($referer = $request->headers->get('referer')) {
        $message .= sprintf(' (from "%s")', $referer);
    }

    throw new NotFoundHttpException($message, $e);
}
```

flip111:
Hello, the RouterListener you linked is from 13th of januari as part of symfony 4.1 development branch (master). Silex matches http-kernel 2.8 or 3.* https://github.com/silexphp/Silex/blob/ec7d5b5334465414952d4b2e935e73bd085dbbbb/composer.json#L22

[...]

PR #1596, opened 2018-02-17
Updated for RouterListener from symfony/http-kernel v3.4+

ProgMiner:
#1591


fabpot:
Thank you @ProgMiner.

PR #1607, opened 2018-02-25
Drop support for older versions of PHP and Symfony

fabpot:
Let's make it much simpler to maintain future versions of Silex. This pull request does the following (@Ocramius would be proud of me :)):

 * drop support for Twig < 2.0
 * drop support for PHP < 7.1
 * drop support for Symfony 2.x and 3.x


hkdobrev:
:raised_hands: just one question - would that be released as a new major version of Silex?

fabpot:
@hkdobrev 2.3.0 would be the version. I don't think we need a new major version here.

doc/changelog.rst
```diff
@@ -4,6 +4,9 @@ Changelog
 2.3.0 (2018-XX-XX)
 ------------------
 
+ * dropped support for Twig < 2.0
+ * dropped support for PHP < 7.0
```
pierredup: I think this should be `< 7.1`

GBeushausen:
I totally agree with all of this, but is supporting only PHP 7.1 and above a little too much? I think we should support 7.0 because this is the version that comes with the latest Ubuntu LTS release wich is widely used on many hosting solutions, i.e. AWS.

fabpot:
@GBeushausen Symfony 4 requires PHP 7.1, so that's why I'm using 7.1+ here as well. And keep in mind that people can still use older Silex versions if they need to.

GBeushausen:
@fabpot Ah ok. Then of course this is the best solution.

Ocramius:
Proud reader here 👍

@fabpot does this effectively mean that you want to keep the project alive under reduced maintenance?

Overall, reducing maintenance burden is a big win anyway, so this is a good idea.

fabpot:
@Ocramius I'm trying to make it easier for a new team to take over :)

PR #1608, opened 2018-02-25
Make CSRF extension work with Validator translations domain

### LINKED ISSUE #1569: [Form] [CSRF] Translation domain not set properly for CsrfExtension
The ```FormServiceProvider``` registers the ```form.extension.csrf``` service as an instance of ```Symfony\Component\Form\Extension\Csrf\CsrfExtension``` class that takes the ```translation_domain``` as third parameter.

The translation domain is passed to ```CsrfValidationListener``` and it used directly for translation. By default, validation errors are registered with the ```validators``` domain (see ```ValidatorServiceProvider```), but the translator has a fallback to the ```messages``` domain, so CSRF errors cannot get translated properly.

**TL;DR** The ```FormServiceProvider``` should check whether the ```ValidatorServiceProvider``` has been registered and use its translation domain as third parameter for the ```CsrfExtension```, null otherwise to allow correct translations of CSRF form errors.

hkdobrev:
- Make Validator translation domain configurable
- Fix #1569.

### DIFF
--- a/src/Silex/Provider/FormServiceProvider.php
+++ b/src/Silex/Provider/FormServiceProvider.php
@@ -47,7 +47,9 @@ class FormServiceProvider implements ServiceProviderInterface
 
         $app['form.extension.csrf'] = function ($app) {
             if (isset($app['translator'])) {
-                return new CsrfExtension($app['csrf.token_manager'], $app['translator']);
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @hkdobrev.

PR #1611, opened 2018-02-25
Allow defining users provider as a service ID

### LINKED ISSUE #1216: SecurityServiceProvider use CallbackResolver

hkdobrev:
Replaces #1217.
Closes #1216.

### DIFF
--- a/doc/providers/security.rst
+++ b/doc/providers/security.rst
@@ -495,8 +495,8 @@ Defining a custom User Provider
 Using an array of users is simple and useful when securing an admin section of
 a personal website, but you can override this default mechanism with you own.
 
-The ``users`` setting can be defined as a service that returns an instance of
-`UserProviderInterface
+The ``users`` setting can be defined as a service or an service id that returns
+an instance of `UserProviderInterface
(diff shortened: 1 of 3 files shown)


doc/providers/security.rst
```diff
@@ -495,8 +495,8 @@ Defining a custom User Provider
 Using an array of users is simple and useful when securing an admin section of
 a personal website, but you can override this default mechanism with you own.
 
-The ``users`` setting can be defined as a service that returns an instance of
-`UserProviderInterface
+The ``users`` setting can be defined as a service or an service id that returns
```
fabpot: typo: `or a service id`

hkdobrev: Thanks, fixed!

fabpot:
Thank you @hkdobrev.

PR #1614, opened 2018-02-25
Use ServiceIterator in Security provider for voters

### LINKED ISSUE #1530: Injecting AccessDecisionManager in to Voter breaks application and kills PHP.
I have a requirement within my application to check user access based on particular roles and attributes, and so it seemed logical to implement a `Voter` to do this.

However, there appears to be an issue when injecting `Symfony\Component\Security\Core\Authorization\AccessDecisionManager` in to a Voter, as per the [documentation](http://symfony.com/doc/2.8/security/voters.html#checking-for-roles-inside-a-voter).

Upon injecting the `AccessDecisionManager` via the `$app['access_manager']` service, the application simply returns `ERR_EMPTY_RESPONSE`.

Furthermore, running the same application via PHP's development server, the server exits with `Segmentation fault: 11`.

```
php -S localhost:8000 web/app.php
PHP 7.0.18 Development Server started at Wed Jun  7 12:46:24 2017
Listening on http://localhost:8000
Document root is /Users/adam/Desktop/silex-voter
Press Ctrl-C to quit.
Segmentation fault: 11
```

I have tested this with `symfony/security` versions `2.8` through `3.3`.

A repository showing the issue is available here:
https://github.com/amustill/silex-voter-issue-demo

[Line 25 or `web/app.php`](https://github.com/amustill/silex-voter-issue-demo/blob/master/web/app.php#L25) is the offender. Commenting out this line and thus removing the dependency injection of `Symfony\Component\Security\Core\Authorization\AccessDecisionManager` makes the application load.

hkdobrev:
Fixes #1530.

### DIFF
--- a/doc/providers/security.rst
+++ b/doc/providers/security.rst
@@ -681,6 +681,32 @@ Symfony `cookbook`_.
     providers. :doc:`How to Create a Custom Authentication System with Guard
     </cookbook/guard_authentication>`
 
+Using Voters to check user permissions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+See the `Security component documentation on voters <http://symfony.com/doc/current/security/voters.html>`_.
+
+By default Silex includes the role hierarchy and authenticated voters.
+If you want to add a custom voter, you need to register it as a service and extend ``security.voter_services``.
+
+.. code-block:: php
+
+    $app['custom_voter'] = function () {
+        return MyCustomVoter();
+    };
+
+    $app->extend('security.voter_services', function (array $voters) {
+        $voters[] = 'custom_voter';
+
+        return $voters;
+    });
+
+.. note::
+    Using the above approach with the service names circular references are avoided
+    and you can use the ``AccessDecisionManager`` in your custom voter
+    to `check for roles inside a voter
+    <http://symfony.com/doc/current/security/voters.html#checking-for-roles-inside-a-voter>`_.
+
 Stateless Authentication
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -12,6 +12,7 @@
 namespace Silex\Provider;
 
 use Pimple\Container;
+use Pimple\ServiceIterator;
 use Pimple\ServiceProviderInterface;
 use Silex\Application;
 use Silex\Api\BootableProviderInterface;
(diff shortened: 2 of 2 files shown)


src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -140,14 +141,29 @@ public function register(Container $app)
...
+
+        $app['security.voter_services'] = function () {
+            return [RoleHierarchyVoter::class, AuthenticatedVoter::class];
         };
 
+        $app[RoleHierarchyVoter::class] = function ($app) {
+            return new RoleHierarchyVoter(new RoleHierarchy($app['security.role_hierarchy']));
+        }
+
+        $app[AuthenticatedVoter::class] = function ($app) {
```
hkdobrev: @fabpot What do you think of using the class names for the service names like that? These are not really for public consumption anyway, but I think it's a good practice for easier migration to the Symfony DIC auto-wiring. Saw it here: https://github.com/opencfp/opencfp/pull/895

doc/providers/security.rst
```diff
@@ -681,6 +681,32 @@ Symfony `cookbook`_.
     providers. :doc:`How to Create a Custom Authentication System with Guard
     </cookbook/guard_authentication>`
 
+Using Voters to check user permissions
```
fabpot: We are using capital letters for title.

doc/providers/security.rst
```diff
@@ -681,6 +681,32 @@ Symfony `cookbook`_.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+See the `Security component documentation on voters <http://symfony.com/doc/current/security/voters.html>`_.
+
+By default Silex includes the role hierarchy and authenticated voters.
+If you want to add a custom voter, you need to register it as a service and extend ``security.voter_services``.
```
fabpot: 80 chars.

doc/providers/security.rst
```diff
@@ -681,6 +681,32 @@ Symfony `cookbook`_.
...
+    };
+
+    $app->extend('security.voter_services', function (array $voters) {
+        $voters[] = 'custom_voter';
+
+        return $voters;
+    });
+
+.. note::
+    Using the above approach with the service names circular references are avoided
```
fabpot: missing blank line above, and comma in the sentence.

fabpot:
Thank you @hkdobrev.

doc/providers/security.rst
fabpot: see a50ba32 for the fixes

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -140,14 +141,29 @@ public function register(Container $app)
...
+        // Unused, kept for backwards-compatibility
+        // Extend security.voter_services instead to prevent circular references
         $app['security.voters'] = function ($app) {
-            return [
-                new RoleHierarchyVoter(new RoleHierarchy($app['security.role_hierarchy'])),
-                new AuthenticatedVoter($app['security.trust_resolver']),
-            ];
+            return array_map(function ($voterServiceId) use ($app) {
+                return $app[$voterServiceId];
+            });
```
stof: this is broken. It misses the second argument of `array_map`.

and it is not BC either: extending `security.voters` to register additional voters is not working anymore, which is a BC break. This is what should preserve BC.

hkdobrev: @stof Thanks for that! Fix on the `array_map` is coming shortly. However, this functionality was not documented before. Should we consider that as a BC break?

hkdobrev: To preserve backwards compatibility we could check whether the list of voter services is the same as the default one and if it is we could use `security.voters` as before.

hkdobrev: @stof Addressed in #1619.

Issue #1577, opened 2017-11-26
TwigServiceProvider assumes twig.options is a fixed array


hkdobrev:
I want the `twig.options` configuration to be a closure and Pimple allows that easily via:

```php
$app->register(new TwigServiceProvider(), [
     'twig.options' => function(Container $app) {
          return [
            'cache' => $app['caching']
                ? $app['cache.dir'].'/twig'
                : false,
          ];
     },
]);
```

However, `TwigServiceProvider` does not allow that, because it is using `array_replace` getting and then setting the `twig.options` key in one go:

https://github.com/silexphp/Silex/blob/c62cfba1b9a3533274ad01b9f55d87000c26071c/src/Silex/Provider/TwigServiceProvider.php#L55-L61

This results in Pimple throwing an exception - which is normal because the provider has just retrieved a key and tried to override it. This overriding works only if it's a fixed value and Pimple has not executed a closure and frozen the result.

hkdobrev:
One good way to address that in my opinion which would make using lazy closures for configuration options on many providers would be to allow extending something which is still a scalar or non-invokable object as it could be overridden to a wrapping closure later before it is accessed.

Old, but related: https://github.com/silexphp/Pimple/issues/70

Then `TwigServiceProvider` would be able to do the `array_replace` operation in an `extend` wrapping closure after the declaration of `$app['twig.options'] = array();`.

This way everything is lazy and it would improve how the provider works even more as it would do the parameter replacement for any Twig factory retrieval and not just for the default `$app['twig']` service like now.

hkdobrev:
@fabpot @skalpa Do you have a suggestion for that problem? I'm keen on tackling that, but I'm not certain I'm on the best path so it's intuitive to use.

skalpa:
Why would you want `twig.options` to be a closure to start with ? What's your use case ? 

If you ask me too many things are already defined with closures, and while it makes sense to register arrays of services like this, I don't believe it does the rest of the time.

Ie: in the pimple issue you linked the solution to the author's problem is just _not_ to use `extend()`:
```php
$c['some_value'] = array(1, 2, 3);
$c['some_value'] = array_merge($c['some_value'], array(4, 5));
```

hkdobrev:
@skalpa I have something like this:

```php
$app->register(new TwigServiceProvider(), [
    'twig.options' => [
        'cache' => $app['caching']
            ? $app['cache.dir'].'/twig'
            : false,
    ],
]);
```

[...]

PR #1616, opened 2018-02-25
Do not re-set Twig options after registering

hkdobrev:
This allows users to have `twig.options` as a closure for lazy evaluation.

Fixes https://github.com/silexphp/Silex/issues/1577.

This is the simplest approach introducing a BC break as `twig.environment_factory` service may have been used to create another Twig instance.

### DIFF
--- a/src/Silex/Provider/TwigServiceProvider.php
+++ b/src/Silex/Provider/TwigServiceProvider.php
@@ -54,15 +54,16 @@ class TwigServiceProvider implements ServiceProviderInterface
         $app['twig.number_format.thousands_separator'] = ',';
 
         $app['twig'] = function ($app) {
-            $app['twig.options'] = array_replace(
+            $options = array_replace(
                 [
                     'charset' => $app['charset'],
                     'debug' => $app['debug'],
                     'strict_variables' => $app['debug'],
-                ], $app['twig.options']
+                ],
+                $app['twig.options']
             );
 
-            $twig = $app['twig.environment_factory']($app);
+            $twig = $app['twig.environment_factory']($options);
             // registered for BC, but should not be used anymore
             // deprecated and should probably be removed in Silex 3.0
             $twig->addGlobal('app', $app);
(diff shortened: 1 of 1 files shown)


src/Silex/Provider/TwigServiceProvider.php
```diff
@@ -54,15 +54,16 @@ public function register(Container $app)
...
         $app['twig'] = function ($app) {
-            $app['twig.options'] = array_replace(
+            $options = array_replace(
                 [
                     'charset' => $app['charset'],
                     'debug' => $app['debug'],
                     'strict_variables' => $app['debug'],
-                ], $app['twig.options']
+                ],
+                $app['twig.options']
```
fabpot: missing comma at the end

src/Silex/Provider/TwigServiceProvider.php
```diff
@@ -54,15 +54,16 @@ public function register(Container $app)
...
             );
 
-            $twig = $app['twig.environment_factory']($app);
+            $twig = $app['twig.environment_factory']($options);
```
fabpot: As you said, this is a BC break, so it cannot be merged without a BC layer. I don't think this is acceptable as I can see several scenarios where one would override this service.

src/Silex/Provider/TwigServiceProvider.php
fabpot: I don't understand how that fixes the issue. Here, if the option is a closure, that will fail.

src/Silex/Provider/TwigServiceProvider.php
hkdobrev: This was just a quick mockup to illustrate my point. I understand it's not mergeable hence my question in https://github.com/silexphp/Silex/issues/1577#issuecomment-368353476

src/Silex/Provider/TwigServiceProvider.php
hkdobrev: Hmm, I don't actually understand how it would fail. I'm only reading a value from the container without overriding it anymore. It was failing before when it was a closure.

> missing comma at the end

This is a function call. Trailing commas in function calls are planned for PHP 7.3.
I've re-formatted it as per PSR-2 which states that when one argument is on its own line, every argument must be on its own line. An alternative formatting is just putting the array elements on their own line:

```php
$options = array_replace([
    'charset' => $app['charset'],
    'debug' => $app['debug'],
    'strict_variables' => $app['debug'],
],$app['twig.options']);
```

PR #1619, opened 2018-02-26
Fix SecurityServiceProviders voters

hkdobrev:
Follow-up to #1614.

See https://github.com/silexphp/Silex/pull/1614#discussion_r170567411

### DIFF
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -164,7 +164,7 @@ class SecurityServiceProvider implements ServiceProviderInterface, EventListener
         $app['security.voters'] = function ($app) {
             return array_map(function ($voterServiceId) use ($app) {
                 return $app[$voterServiceId];
-            });
+            }, $app['security.voter_services']);
         };
 
         $app['security.firewall'] = function ($app) {
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -12,9 +12,9 @@
 namespace Silex\Provider;
 
 use Pimple\Container;
-use Pimple\ServiceIterator;
 use Pimple\ServiceProviderInterface;
 use Silex\Application;
+use Silex\ServiceIterator;
 use Silex\Api\BootableProviderInterface;
 use Silex\Api\ControllerProviderInterface;
 use Silex\Api\EventListenerProviderInterface;
@@ -141,22 +141,34 @@ class SecurityServiceProvider implements ServiceProviderInterface, EventListener
             return new UserChecker();
         };
 
-        $defaultVoters = [RoleHierarchyVoter::class, AuthenticatedVoter::class];
+        $defaultVoterServiceIds = [RoleHierarchyVoter::class, AuthenticatedVoter::class];
 
-        $app['security.access_manager'] = function ($app) use ($defaultVoters) {
-            $voter_services = $app['security.voter_services'];
-            if (empty(array_diff($voter_services, $defaultVoters))) {
-                // BC: If you don't need lazy voter service evaluation, you could still extend security.voters
-                $voters = $app['security.voters'];
-            } else {
-                $voters = new ServiceIterator($app, $voter_services);
-            }
-
-            return new AccessDecisionManager($voters);
+        $app['security.__default_voters'] = function () use ($defaultVoterServiceIds) {
+            return array_map(function ($voterServiceId) use ($app) {
+                return $app[$voterServiceId];
+            }, $defaultVoterServiceIds);
         };
 
-        $app['security.voter_services'] = function () use ($defaultVoters) {
-            return $defaultVoters;
+        $app['security.access_manager'] = function ($app) use ($defaultVoterServiceIds) {
+            $voters = [];
+
+            if ($app['security.voters'] !== $app['security.__default_voters']) {
+                @trigger_error(
+                    'Using security.voters is deprecated! Extend security.voter_service_ids instead',
+                    E_USER_DEPRECATED
+                );
+                $voters = $app['security.voters'];
+            }
+
+            if ($app['security.voter_service_ids'] !== $defaultVoterServiceIds || empty($voters)) {
+                $voters = array_merge($voters, $app['security.voter_service_ids']);
+            }
+
+            return new AccessDecisionManager(new ServiceIterator($app, $voters));
+        };
+
+        $app['security.voter_service_ids'] = function () {
+            return $defaultVoterServiceIds;
         };
 
         $app[RoleHierarchyVoter::class] = function ($app) {
@@ -168,11 +180,9 @@ class SecurityServiceProvider implements ServiceProviderInterface, EventListener
         };
 
         // Unused, kept for backwards-compatibility
-        // Extend security.voter_services instead to prevent circular references
+        // Extend security.voter_service_ids instead to prevent circular references
         $app['security.voters'] = function ($app) {
-            return array_map(function ($voterServiceId) use ($app) {
-                return $app[$voterServiceId];
-            }, $app['security.voter_services']);
+            return $app['security.__default_voters'];
         };
 
         $app['security.firewall'] = function ($app) {
--- a/src/Silex/Provider/SecurityServiceProvider.php
+++ b/src/Silex/Provider/SecurityServiceProvider.php
@@ -86,8 +86,6 @@ class SecurityServiceProvider implements ServiceProviderInterface, EventListener
         $app['security.hide_user_not_found'] = true;
         $app['security.encoder.bcrypt.cost'] = 13;
 
-        $default_voters = [RoleHierarchyVoter::class, AuthenticatedVoter::class];
-
         $app['security.authorization_checker'] = function ($app) {
             return new AuthorizationChecker($app['security.token_storage'], $app['security.authentication_manager'], $app['security.access_manager']);
         };
@@ -143,9 +141,11 @@ class SecurityServiceProvider implements ServiceProviderInterface, EventListener
             return new UserChecker();
         };
 
-        $app['security.access_manager'] = function ($app) use ($default_voters) {
+        $defaultVoters = [RoleHierarchyVoter::class, AuthenticatedVoter::class];
+
+        $app['security.access_manager'] = function ($app) use ($defaultVoters) {
             $voter_services = $app['security.voter_services'];
-            if (empty(array_diff($voter_services, $default_voters))) {
+            if (empty(array_diff($voter_services, $defaultVoters))) {
                 // BC: If you don't need lazy voter service evaluation, you could still extend security.voters
                 $voters = $app['security.voters'];
             } else {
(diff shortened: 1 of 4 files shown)


src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -85,6 +85,8 @@ public function register(Container $app)
         $app['security.hide_user_not_found'] = true;
         $app['security.encoder.bcrypt.cost'] = 13;
 
+        $default_voters = [RoleHierarchyVoter::class, AuthenticatedVoter::class];
```
stof: please use camelCased variable names

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -140,14 +142,20 @@ public function register(Container $app)
             return new UserChecker();
         };
 
-        $app['security.access_manager'] = function ($app) {
-            $votersIterator = new ServiceIterator($app, $app['security.voter_services']);
+        $app['security.access_manager'] = function ($app) use ($default_voters) {
+            $voter_services = $app['security.voter_services'];
+            if (empty(array_diff($voter_services, $default_voters))) {
+                // BC: If you don't need lazy voter service evaluation, you could still extend security.voters
```
stof: we need to trigger a deprecation warning if they extend stuff in the old way rather than the new one, to tell them to migrate

src/Silex/Provider/SecurityServiceProvider.php
```diff
@@ -140,14 +142,20 @@ public function register(Container $app)
             return new UserChecker();
         };
 
-        $app['security.access_manager'] = function ($app) {
-            $votersIterator = new ServiceIterator($app, $app['security.voter_services']);
+        $app['security.access_manager'] = function ($app) use ($default_voters) {
+            $voter_services = $app['security.voter_services'];
+            if (empty(array_diff($voter_services, $default_voters))) {
```
stof: what we actually need to know is whether `$app['security.voters']` was extended to register some in the old way, not whether `$app['security.voter_services']` was extended.

If you extend both, we still need to register the ones registered in the old way, which your code does not.
And the default implementation of `$app['security.voters']` means that any voter registered in the new way would be available in it if needed.

src/Silex/Provider/SecurityServiceProvider.php
hkdobrev: @stof I'm not sure extending `security.voters` should be deprecated though unless we want to move to lazy-only. I mean you may not have a circular reference dependency in your voter.

src/Silex/Provider/SecurityServiceProvider.php
hkdobrev: A simpler implementation could be achieved if the `Pimple\ServiceIterator` class accepts either service ID or instances. A bit similar to how Doctrine listener resolvers work http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html#entity-listeners-resolver

src/Silex/Provider/SecurityServiceProvider.php
hkdobrev: Fixed.

hkdobrev:
@stof I realise where you're coming from, but I think we could have a discussion whether that should be a deprecation here and that should be merged as a fix as the current master is broken after #1614 :( This way more things could be released as patch versions. What do you think?

src/Silex/Provider/SecurityServiceProvider.php
stof: Well, having 2 different ways to register voters (especially when they cannot work in parallel, as was the case at the time of my previous review) is definitely a bad DX.

stof:
This PR is still broken. If you extend both the `security.voter_services` and `security.voters` arrays (which could happen in different providers), only extensions of the `security.voter_services` service are taken into account, as explained in my previous review.

Breaking code using the old extension as soon as *any* provider starts using the new one means that the BC layer is broken (as soon as someone starts migrating, there is no BC layer anymore).

src/Silex/Provider/SecurityServiceProvider.php
hkdobrev: We can make it work in parallel if we register services on the fly based on the voters objects and extend the services with them, but this seems quite hacky.

hkdobrev:
I have 2 suggestions to fix it so far:

> We can make it work in parallel if we register services on the fly based on the voters objects and extend the services with them, but this seems quite hacky.

> A simpler implementation could be achieved if the Pimple\ServiceIterator class accepts both service ID and instances

hkdobrev:
@fabpot Would you have a moment to chime in on this? How do you think BC should be handled here? What's actually a breaking change?

skalpa:
You can solve the BC problem by creating a custom iterator that is able to iterate over an array of service ids *and* an array of objects in sequence.

```php
namespace Silex;

/** @internal **/
final class ServiceIterator implements \Iterator
{
    private $iterated;
    private $idsCount;
    private $index = 0;

    public function __construct(Application $app, array $serviceIds, array $serviceObjects)
    {
        $this->idsCount = count($serviceIds);
        $this->iterated = array_merge(array_values($serviceIds), array_values($serviceObjects));
    }

    public function current()
    {
        $current = $this->iterated[$this->index++];

        return $this->index > $this->idsCount ? $current : $this->app[$current];
    }
    // .......
}
```

Then make sure users get a deprecation notice if `$app['security_voters']` is not empty, and inject `new ServiceIterator($app, $app['security.voter_services'], $app['security.voters'])`.

At a later stage you'll be able to get rid of `security.voters` and inject a plain `Pimple\ServiceIterator` instance in place of your `Silex\ServiceIterator`.

Another remark: `security.voter_services` should be renamed to `security.voter_service_ids` as it is not an array of services, but an array of service ids (and that would be consistent with `validator.validator_service_ids` for instance).

hkdobrev:
> You can solve the BC problem by creating a custom iterator that is able to iterate over an array of service ids and an array of objects in sequence.

👍 That's what I've suggested above in https://github.com/silexphp/Silex/pull/1619#discussion_r170618144

> At a later stage you'll be able to get rid of security.voters and inject a plain Pimple\ServiceIterator instance in place of your Silex\ServiceIterator.

Do you mean that we should implement the service iterator in Silex, but later move the implementation to Pimple when we upgrade to require minimum next minor release of Pimple?

> Another remark: `security.voter_services` should be renamed to `security.voter_service_ids` as it is not an array of services, but an array of service ids (and that would be consistent with `validator.validator_service_ids` for instance).

👍 Totally with you here, couldn't find what the convention for these is.

hkdobrev:
> Then make sure users get a deprecation notice if `$app['security_voters']` is not empty

That's not that easy unfortunately. To preserve BC we need this to continue to work:
```php
$app->extend('security.voters`, function(array $voters) {
	$voters[] = new MyCustomVoter();

	return $voters;
});
```

And we need to keep the two default voters inside that voters array. We can easily check if `security.voter_service_ids` was modified as we can just diff the strings inside. For the voter services we may end up instantiating the default voters even if the users have used the new way and overridden the defaults.

skalpa:
No, I don't think this belong in Pimple. I mean just get rid of it because it won't be needed it anymore:

* Tell users that using `security.voters` is deprecated and that it will be removed in the next major version of Silex
* In the next major Silex, remove `security.voters`
* At that point you'll just need an iterator that can `new ServiceIterator($app, $app['security.voter_service_ids'])`, and that's exactly how the Pimple iterator works. Thus you'll be able to remove the temporary `Silex\ServiceIterator` and use the Pimple class instead (note that to be able to do this and just remove the class you have to add a docblock to mark it as `@internal` so users know they should not rely on it).

hkdobrev:
On the service container, I've done it like you describe, I just thought Pimple could use that capability in general. But anyway, this is not the blocker here. What do you think about my BC comment?

skalpa:
Regarding BC, I just realized that what I advised covered the _users can add voters to the defaults_ use case but not __users can replace the list of voters completely__ . 

If I'm correct, this should be 100% BC instead:
```
$app['security.voter_service_ids'] = function () { return array(); };
$app['security.__default_voters'] = function () {
    return array(new Voter1(), new Voter2());
}; 
$app['security.voters'] = function () { return $app['security.__default_voters']; };

$app['security.access_manager'] = function ($app) {
    if ($app['security.voters'] !== $app['security.__default_voters']) {
        @trigger_error('Blah', E_USER_DEPRECATED);
    }

    return new AccessDecisionManager(new ServiceIterator($app, $app['security.voter_service_ids'], $app['security.voters']));
};
```

hkdobrev:
Let's separate all cases:

1. User hasn't modified any voters and access the `security.access_manager` service with the defaults.

- `ServiceIterator` is passed to `AccessDecisionManager` with service IDs only
- No deprecation warning is triggered
- `RoleHierarchyVoter` and `AuthenticatedVoter` voters are used

2. User has modified the `security.voters` service, but has not modified the `security.voter_service_ids`.

- `ServiceIterator` is passed to `AccessDecisionManager` with instances only
- Deprecation warning is triggered
- The modified voters are used.


3. User has modified the `security.voter_service_ids` service, but has not modified the `security.voters`.

- `ServiceIterator` is passed to `AccessDecisionManager` with service IDs only
- No deprecation warning is triggered
- The modified voters are used lazily.

4. User has modified both the `security.voter_service_ids` and `security.voters` services.

- `ServiceIterator` is passed to `AccessDecisionManager` with both service IDs and instances
- Deprecation warning is triggered
- All of the modified voters are used

hkdobrev:
I've implemented the `ServiceIterator` supporting both cases in one argument as I think it's more useful for its purpose.

fabpot:
@hkdobrev You need to get rid of the merge commit. We do not merge pull request with merge commit. Instead, you can rebase on current master (that will automatically remove the merge commit).

hkdobrev:
@fabpot Done. Thought I'd use the GitHub conflict resolution for once as it was just for the changelog, but unfortunately it doesn't support rebasing.

On a related note, I'll send a PR telling Git to choose both sides when there's a conflict from additions in the changelog reducing conflicts at least locally via `.gitattributes`.

hkdobrev:
✅ Fixed tests
❌ Triggered deprecation warning
📝 Updated changelog
📜 Updated docs

@stof @skalpa @fabpot Any input?

skalpa:
Your latest version still breaks BC. Users are currently able to remove the default voters completely by using:
```
$app->extend('security.voters', function () {
    return array();
});
```
But that wouldn't work anymore...

Also, the way default voters are added to both `security.voters` and `security.voter_service_ids` is confusing and will lead to the registration of the same voter multiple times if custom voters are added to both parameters (ie: if a users chooses to use `security.voter_service_ids` while a 3rd-party service providers adds its voters to `security.voters`).

Honestly, while I obviously see the benefits of using iterators more globally, I think this hasn't been thought through well enough, and I'd recommend to just revert #1614 and fix master for the time being, so a better solution can be designed later. 

The circular dependencies problem it was supposed to fix can be avoided easily anyway, so there's no rush.

hkdobrev:
> The circular dependencies problem it was supposed to fix can be avoided easily anyway

How could it be avoided?

src/Silex/ServiceIterator.php
```diff
@@ -0,0 +1,67 @@
...
+        if (\is_string($item) && isset($this->container[$item])) {
+            return $this->container[$item];
+        }
+
+        return $item;
+    }
+
+    public function key()
+    {
+        return \current($this->serviceItems);
```
xabbuh: Does it really make sense that this method can return a key or a value?

skalpa:
You can avoid circular dependencies by injecting the application instead of the AccessDecisionManager, as [per my comment in the original issue](https://github.com/silexphp/Silex/issues/1530#issuecomment-306795917).

hkdobrev:
Closing in favour of #1634.

PR #1625, opened 2018-03-12
Support PSR-3 log levels in MonologServiceProvider

hkdobrev:
Monolog already support them everywhere, but when using the `MonologServiceProvider` you cannot use them reliably.


fabpot:
Thank you @hkdobrev.

PR #1629, opened 2018-03-17
Expose AuthenticationUtils

### LINKED ISSUE #1620: Last security error is not translated
Found this question on StackOverflow: [Silex, how to translate login error messages?
](https://stackoverflow.com/questions/48005062/silex-how-to-translate-login-error-messages)

Looking the code it seems indeed the error message is not being translated:

https://github.com/silexphp/Silex/blob/e41379abc3e943fd5ff08e65158decca7af5748d/src/Silex/Provider/SecurityServiceProvider.php#L415-L429

Should we translate it if the translator is registered in the `'security.last_error'` callback or should we expose the `AuthenticationUtils` class in the controllers like Symfony? I don't see translation of the message there as well.

Any pointers?

hkdobrev:
Close #1620.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -15,6 +15,7 @@ Changelog
  * dropped support for Symfony 2.x and 3.x
  * added support for Symfony 4
  * added support PSR-3 log levels in MonologServiceProvider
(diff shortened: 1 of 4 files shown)


fabpot:
Thank you @hkdobrev.

Issue #1630, opened 2018-03-20
The #1611 patch does not treat the user provider as a service


skalpa:
@henrikbjorn @hkdobrev To create an alias to an already registered service you have to use a closure that accesses the service and cannot just copy the factory with `raw()`. Failure to do so may result in the instanciation of your object twice.

In the new test, this will fail while it should work: 
`$this->assertSame($app['my_user_provider'], $app['security.user_provider.default']);`

To fix that, you need to replace `$users = $app->raw($users)` in the provider with:
```
$users = function () use ($app, $users) {
    return $app[$users];
};
```

PS: Apologies for not sending a PR, I'm barely able to find free time for reviews.

PR #1631, opened 2018-03-22
Treat user provider as a service in security service provider

hkdobrev:
Follow-up to https://github.com/silexphp/Silex/pull/1611 and https://github.com/silexphp/Silex/pull/1217.
Fixes https://github.com/silexphp/Silex/issues/1630.


fabpot:
Thank you @hkdobrev.

PR #1632, opened 2018-03-22
Replace twig.options defaults every time

hkdobrev:
Fixes https://github.com/silexphp/Silex/issues/1577
Alternative to #1616.


fabpot:
Thank you @hkdobrev.

PR #1636, opened 2018-03-31
Remove help wanted text from README

fabpot:
Despite some great work done in the last couple of months, nobody showed a strong motivation to take over this project.

I want to thank @hkdobrev for his hard work as he helped me lot in closing issues and fixing bugs. Silex is now compatible with Symfony 4 and that's going to help a lot in migrating projects.

I still strongly believe that Symfony 4 is a better alternative and that there is nothing unique anymore in Silex.

Forking is still an option if someone wants to take over of course.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,8 +3,6 @@ Silex, a simple Web Framework
 
 **WARNING**: Silex is in maintenance mode only. Ends of life is set to June
 2018. Read more on `Symfony's blog <http://symfony.com/blog/the-end-of-silex>`_.
-But Silex can be saved! Any **help** to keep Silex alive is greatly appreciated.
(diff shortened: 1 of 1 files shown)


agilare:
Thank you for this work, which helps to look forward for existing Silex projects.

Then, after 2 months of "help wanted" announcement, neither the current contributors, nore other people wanted to take over Silex ? On reason would be, that there is no reason now to start or continue a project with Silex, because, like you say, Symfony 4 is a better way to follow.

It would be nice to create a short guide explaining how to migrate from Silex to Symfony 4 like  https://github.com/symfony/symfony/blob/master/UPGRADE-4.0.md

Now that end of Silex seems clear the [homepage of the framework](https://silex.symfony.com/) should be updated.

fabpot:
The website now has a warning as well.

PR #1638, opened 2018-04-04
Simplify extend call

hkdobrev:
Since Pimple 2.x there is no need to re-assign the result from `extend()` call to the same key. Pimple already does that.


fabpot:
Thank you @hkdobrev.

Issue #1640, opened 2018-04-23
context logic


719media:
This is concerning the issue #1568
and associated commit at
https://github.com/silexphp/Silex/commit/d92d5c5a1d8bf70b676f8eba10316ac16ee07eab#diff-04d7a599b7a1b46b840ac8be261de1fc

If I have firewalls such as
```
'security.firewalls' => [
  'alpha' => [
    'context' => 'gamma'
  ],
  'beta' => [
    'context' => 'gamma'
  ]
]
```

I still get an `UnknownIdentifierException` exception, only now it is for the firewall name (`beta`) instead of the context. Seems that the !isset check on the user_provider.$name should just be moed outside of the context_listener if block?

Thank you for your good work and service

hkdobrev:
Hi @719media! Which version of Silex are you using? Have you tried latest master?

It seems you already have a suggestion in place. Do you mind opening a pull request? Much appreciated!

719media:
This error happens when using Silex 2.3
PR created in #1641.

PR #1641, opened 2018-04-23
Set new security.user_provider inside existing context

719media:
Set `$app['security.user_provider.'.$name]` even if the context already exists. See #1640


fabpot:
Thank you @719media.

